PR #1, opened 2018-10-05
Fix ENOENT on Windows

marrony:


src/lib/misc.js
```diff
@@ -206,7 +206,7 @@ function writeFile(fileName, data, mode) {
 * Tests if an error is of the type "file not found".
 */
 function fileNotFound(err) {
-  return err.code === 'ENOENT' && err.syscall === 'open' && err.errno === -2
+  return err.code === 'ENOENT' && err.syscall === 'open'
```
marrony: windows returns `errno = -4058`

src/lib/misc.js
```diff
@@ -206,7 +206,7 @@ function writeFile(fileName, data, mode) {
 * Tests if an error is of the type "file not found".
 */
 function fileNotFound(err) {
-  return err.code === 'ENOENT' && err.syscall === 'open' && err.errno === -2
```
bitbckt: Is the implication here that `code` is portable, but `errno` isn't?

PR #6, opened 2019-05-15
SHEL-24: add-endpoint should accept alias and key flags

trevorsibanda:
Adds support for 
`fauna add-endpoint https://localhost:8443/ --alias localhost --key secret`

### DIFF
--- a/src/commands/add-endpoint.js
+++ b/src/commands/add-endpoint.js
@@ -1,4 +1,5 @@
 const {cli} = require('cli-ux')
+const { flags } = require('@oclif/command')
 const {saveEndpointOrError, errorOut} = require('../lib/misc.js')
 const FaunaCommand = require('../lib/fauna-command.js')
 const url = require('url')
@@ -6,6 +7,8 @@ const url = require('url')
 class AddEndpointCommand extends FaunaCommand {
   async run() {
     const endpoint = this.args.endpoint
+    let secret = this.flags.key
+    let alias = this.flags.alias
     const log = this.log
 
     const newEndpoint = url.parse(endpoint)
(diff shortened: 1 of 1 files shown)


src/commands/add-endpoint.js
```diff
@@ -1,32 +1,32 @@
...
+const { flags } = require('@oclif/command')
 const {saveEndpointOrError, errorOut} = require('../lib/misc.js')
 const FaunaCommand = require('../lib/fauna-command.js')
 const url = require('url')
 
 class AddEndpointCommand extends FaunaCommand {
   async run() {
     const endpoint = this.args.endpoint
+    var secret = this.flags.key
+    var alias = this.flags.alias
```
BrunoQuaresma: instead of var you can use `let` just to follow the es6 pattern.

src/commands/add-endpoint.js
```diff
@@ -1,32 +1,32 @@
...
-    .catch(function (err) {
+    if (!secret)
+      secret = await cli.prompt('Endpoint Key', {type: 'hide', timeout: 120000})
+    if (!alias)
+      alias = await cli.prompt('Endpoint Alias', { default: newEndpoint.hostname, timeout: 120000 })
+
+    if (alias === 'default' || alias === 'cloud') {
+      throw new Error(`The word '${alias}' cannot be used as an alias.`)
+    }
+    return saveEndpointOrError(newEndpoint, alias, secret).then(function (_) {
```
BrunoQuaresma: You can remove the _(underline) variable if you are not using it.

BrunoQuaresma: I don't know what is the pattern for async calls on fauna-shell but you can use the new `async/wait` syntax instead of promises if other places on codebase are using it.

PR #10, opened 2019-06-24
SHEL-8: Deprecate run-queries and add eval command

trevorsibanda:
Adds support for
```bash
  $ fauna eval dbname "Paginate(Classes())"

  $ fauna eval dbname --file=/path/to/queries.fql

  $ echo "Add(1,1)" | fauna eval dbname --stdin

  $ fauna eval dbname "Add(2,3)" --output="/tmp/result"

  $ fauna eval dbname "Add(2,3)" --format=json --output="/tmp/result"
```

### DIFF
--- a/src/commands/eval.js
+++ b/src/commands/eval.js
@@ -1,0 +1,170 @@
+const util = require('util')
+const fs = require('fs')
+const esprima = require('esprima')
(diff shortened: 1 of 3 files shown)


bitbckt:
I don't know anything about this.

PR #15, opened 2019-07-23
SHEL-30: Use fixed version of faunadb driver

marrony:
I don't know the pros and cons of using a fixed version, but given our versioning doesn't follow semver I think doesn't make sense to leave it open.

The problem with this is that, devs using `fauna/faunadb:2.6.5` to test, if they install fauna-shell freshly, it will resolve the driver dependency to `faunadb@2.7.0` which is incompatible

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "cli-ux": "^4.8.0",
     "escodegen": "^1.11.0",
     "esprima": "^4.0.1",
-    "faunadb": "^2.5.1",
+    "faunadb": "~2.5.1",
     "globby": "8",
     "heroku-cli-util": "^8.0.9",
     "ini": "^1.3.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "cli-ux": "^4.8.0",
     "escodegen": "^1.11.0",
     "esprima": "^4.0.1",
-    "faunadb": "^2.5.1",
+    "faunadb": "2.5.1",
```
marrony: I was suggested to use tilde instead of caret

BrunoQuaresma:
@marrony instead of use the 2.5.1 version why not 2.7.0?

marrony:
I wanted to create at least one version compatible with fauna 2.6.5 and then update to 2.7.0

PR #21, opened 2019-10-16
Strip login via github from shell until it works

benjumanji:


benjumanji:
Closed in favour of friendlier fix.

PR #22, opened 2019-10-16
CON-639 Remove --github from cloud login command

BrunoQuaresma:


src/commands/cloud-login.js
```diff
@@ -1,41 +1,54 @@
...
       errorOut(err.message, 1)
     })
   }
 }
 
-async function cloudStrategy() {
-  const email = await cli.prompt('Email', {timeout: 120000})
+async function cloudStrategy({log}) {
+  log('For email login, enter your email below, and then your password.')
+  log('For login from with 3rd-party identity providers like Github or Netlify, please acquire a key from Dashboard > Security and enter it below instead.')
```
fauna-ryan: "for login from with" -> "for login with"

danieltodonnell: Fix this text and we're good to merge

PR #31, opened 2020-01-22
FE-70: Fix fauna eval by removing the required db-name and prints correct output format

marrony:
this PR allow users to integrate fauna eval with system commands like `jq`, for example:
```bash
key=$(fauna eval --secret=secret 'CreateKey({role: "admin"})' | jq '.[0].secret')

fauna eval --secret=secret 'If(Exists(Database("db")), Delete(Database("db")), null)'

fauna eval --secret=secret 'CreateDatabase({name: "db"})'

fauna eval --secret="$key:db:admin" 'CreateCollection({name: "coll"})'

fauna eval --secret="$key:db:admin" 'CreateRole({name: "aRole", privileges: {resource: Collection("coll"), actions: {create: false}}})'
```

fyi: I already had that modification running locally for a long time, I just decided to make it a official PR

### DIFF
--- a/src/commands/eval.js
+++ b/src/commands/eval.js
@@ -7,7 +7,7 @@ const {readFile, runQueries, errorOut, loadEndpoints, writeFile} = require('../l
 const faunadb = require('faunadb')
 const q = faunadb.query
 
-const EVAL_OUTPUT_FORMATS = ['json']
+const EVAL_OUTPUT_FORMATS = ['json', 'shell']
 
 function infoMessage(err) {
   const fe = util.inspect(err.faunaError, {depth: null})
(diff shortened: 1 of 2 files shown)


src/commands/eval.js
```diff
@@ -127,11 +134,11 @@ Output format can be specified.
 `
 
 EvalCommand.examples = [
-  '$ fauna eval dbname "Paginate(Classes())"',
-  '$ fauna eval dbname --file=/path/to/queries.fql',
-  '$ echo "Add(1,1)" | fauna eval dbname --stdin',
-  '$ fauna eval dbname "Add(2,3)" --output=/tmp/result"',
-  '$ fauna eval dbname "Add(2,3)" --format=json --output=/tmp/result"',
+  '$ fauna eval "Paginate(Classes())"',
```
fauna-ryanharris: @marrony Would it be appropriate for us to also add this command and its examples to the repo's `README`?

marrony: added

evbruno:
praised be!

PR #32, opened 2020-01-23
Rremove quotes on commands (to make copy/paste easier)

fauna-rodrigo:
- https://faunadb.atlassian.net/browse/FE-251

# How to test
- Run create-database command
- Check if the given examples to open shell are not surrounded by single OR double quotes 

# Screenshots
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,11 +53,11 @@ created database 'my_app'
 
 To start a shell with your new database, run:
 
-	fauna shell 'my_app'
+	fauna shell my_app
 
 Or, to create an application key for your database, run:
 
-	fauna create-key 'my_app'
+	fauna create-key my_app
 ```
 
 And then listing your databases:
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ This is how you can create a database called `my_app`:
 $ fauna create-database my_app
 creating database my_app
 
-created database 'my_app'
+created database my_app
 
 To start a shell with your new database, run:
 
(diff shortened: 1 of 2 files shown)


fauna-ryanharris:
@rodrigomaia17 I just had two quick questions:

1. The ticket says to replace the single quotes with double quotes, but we have removed them here. Is that our intention (just asking for my own clarification in case I missed something)?

2. For consistency's sake, should we also update the following log line?
```
created database 'test-cli3'
```

fauna-rodrigo:
> @rodrigomaia17 I just had two quick questions:
> 
> 1. The ticket says to replace the single quotes with double quotes, but we have removed them here. Is that our intention (just asking for my own clarification in case I missed something)?
> 2. For consistency's sake, should we also update the following log line?
> 
> ```
> created database 'test-cli3'
> ```

@fauna-ryanharris 1. Yeah we opted to remove them altogether (since it is the pattern used in all other commands)., I ve added  a comment in the ticked but missed here. thank you
2 . Sure! makes sense! I will updated it now.

fauna-ryanharris:
> @fauna-ryanharris 1. Yeah we opted to remove them altogether (since it is the pattern used in all other commands)., I ve added a comment in the ticked but missed here. thank you
> 2 . Sure! makes sense! I will updated it now.

@rodrigomaia17 Thanks for the context! Appreciate you clarifying 🚀

PR #33, opened 2020-01-23
FE-249 - remove https for add-endpoint localhost example

fauna-rodrigo:
- https://faunadb.atlassian.net/browse/FE-249

# How to test
- Run `fauna add-endpoint --help` 
- Check if the localhost example use http instead of https. 
- If you have a localhost instance running, you should be able to copy/paste that example and run it flawlessly.

# Screenshot
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,7 @@ DESCRIPTION
 
 EXAMPLE
   $ fauna add-endpoint https://db.fauna.com:443
+  $ fauna add-endpoint http://localhost:8443/ --alias localhost --key secret
 ```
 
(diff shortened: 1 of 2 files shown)


fauna-rodrigo:
> Good catch! I ran into this issue setting up my environment last week.
> 
> What do you think about adding this example to the `README` file, as well? Right now, we only have a HTTPS example on line 422.

@fauna-ryanharris Great catch! I ve added the same line to our readme following your suggestion. let me know what you think. 

Thank you for the great suggestion :tada:

fauna-ryanharris:
@fauna-rodrigo Works for me! Thanks for updating 🙏

PR #36, opened 2020-02-24
FE-306 | Fix bug where shell only returns results from last FQL expression

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-306)

### How to test
1. Navigate into the directory for your local `fauna-shell` repo
2. Connect to a local DB endpoint using the executable in the `bin/` directory

```bash
$ bin/run shell --endpoint localhost_hi
```
3. Execute queries with multiple expressions delimited by semicolons. For example:

```
Paginate(Collections());
Paginate(Indexes());
```

4. You should see the results from ALL the FQL expressions, not just the last one

**Updated**
5. Also try this query, which @trevorsibanda discovered wasn't outputting the proper error message. Now, it should work properly:

```
[CreateCollection({name: "orders"}), CreateIndex({name: "idx", source: Collection("x2")})]
```
### Screenshots
image

### DIFF
--- a/src/commands/shell.js
+++ b/src/commands/shell.js
@@ -1,9 +1,10 @@
 const FaunaCommand = require('../lib/fauna-command.js')
-const {errorOut} = require('../lib/misc.js')
+const { errorOut, runQueries } = require('../lib/misc.js')
 const faunadb = require('faunadb')
 const q = faunadb.query
 const repl = require('repl')
 const util = require('util')
+const esprima = require('esprima')
 
 /**
 * We need this function to allow multi-line javascript objects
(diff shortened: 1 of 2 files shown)


src/commands/shell.js
```diff
@@ -68,25 +69,32 @@ function startShell(client, endpoint, dbscope, log) {
...
         })
-        .catch(function (error) {
+        .catch(error => {
           ctx.lastError = error
-          log('Error:', error.message)
+          log('Error:', error.faunaError.message)
+          console.log(util.inspect(JSON.parse(error.faunaError.requestResult.responseRaw), {
+            depth: null,
+            compact: false
+          }))
```
danieltodonnell: Did you mean to leave this console.log in?

BrunoQuaresma: It looks a missed debug thing 👀

fauna-ryanharris: @danieltodonnell @BrunoQuaresma This is what logs out the details of the error (highlighted in the screenshot below). Line 93 simple logs out `Error: error message`, but not the details. This is similar to what we do on line 88 for successful queries.

75267441-e85dea00-57fe-11ea-9003-7cfd0e83c258

PR #58, opened 2020-08-18
FE-736 | Update to driver 3.0.1

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-736)

This PR updates the JS Driver to v3.0.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1597,12 +1597,11 @@
       }
     },
     "cross-fetch": {
-      "version": "3.0.4",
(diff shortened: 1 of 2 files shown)


fauna-ryanharris:
> Add new commands

@BrunoQuaresma I don't see a commands array like in Dashboard and VS Code 🤔

BrunoQuaresma:
Interesting, we could take a look on this after to see how fauna-shelll use queries without include them explicitly.

PR #65, opened 2020-10-13
FE-818 | Prompt user for OTP when MFA is enabled on account

ryancharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-818)

Now that we've enabled MFA in Dashboard, we want to make sure we're protecting data that could be accessed via Fauna Shell. Whenever a user runs `cloud-login` but has MFA enabled on their account, they will now be prompted to enter their OTP code before successfully authenticating.

### How to test
~~1. Pull down this branch
2. Set up a user on https://dashboard.fauna-preview.com/ and enable MFA
3. Back in your local shell, run `bin/run cloud-login`
4. You should be prompted for a username, password and an OTP code
5. Upon success, run `bin/run shell --endpoint [NAME]` where `name` is whatever you called the endpoint in Step 3
6. Run some FQL queries to confirm everything is working~~

### How to test locally
1. Run your local `dashboard`, `auth-service` and `db`
2. Login to the Dashboard, create a new user, enable MFA and scan the QR code
3. In `cloud-login.js`, you'll want to make the changes below:
```javascript
const SHELL_LOGIN_URL = process.env.FAUNA_SHELL_LOGIN_URL
  ? process.env.FAUNA_SHELL_LOGIN_URL
  : "http://localhost:3000/login";

const CLOUD_URL = "http://localhost:8443";
```
4. In the repo directory, run `bin/run cloud-login`. You should be prompted for a email, password and OTP code (since you enabled MFA in step 2)
5. Now, run `bin/run shell --endpoint cloud` and run a few queries to confirm everything is working

### Screencap
https://www.loom.com/share/183072399cbb4e6890701fb97e627ace


src/lib/misc.js
```diff
@@ -10,7 +10,7 @@ const escodegen = require('escodegen')
 const Errors = require('@oclif/errors')
 var rp = require('request-promise')
 
-const FAUNA_CLOUD_DOMAIN = 'db.fauna.com'
+const FAUNA_CLOUD_DOMAIN = 'db.fauna-preview.com'
```
ryancharris, line 13: When running `cloud-login`, the `validCloudEndpoint()` function checks to see if this string is in the URL. If not, it logs the error on line 15. The error doesn't show up the first time you configure the `cloud` endpoint, but will be displayed any subsequent time `cloud-login` is run and the endpoint setup will fail.

BrunoQuaresma (commented):
Do you think we can do better with the URLs? Don't look a good thing have these values hard coded.

ryancharris:
> Do you think we can do better with the URLs? Don't look a good thing have these values hard coded.

@BrunoQuaresma They were hardcoded in the past, so I just updated the values. Should I move these to a `.env` file?

BrunoQuaresma:
> > Do you think we can do better with the URLs? Don't look a good thing have these values hard coded.
> 
> @BrunoQuaresma They were hardcoded in the past, so I just updated the values. Should I move these to a `.env` file?

Maybe, we just have to remember to add these to the CI/CD pipeline. My fear is to deploy a version of fauna-shell using preview... which is going to break fauna-shell at all.

rodrimaia:
> > > Do you think we can do better with the URLs? Don't look a good thing have these values hard coded.
> > 
> > 
> > @BrunoQuaresma They were hardcoded in the past, so I just updated the values. Should I move these to a `.env` file?
> 
> Maybe, we just have to remember to add these to the CI/CD pipeline. My fear is to deploy a version of fauna-shell using preview... which is going to break fauna-shell at all.

Maybe we can make it a way that uses the local environment variable (if it exists) OR the harcoded production URL. (and maybe disclose this in the shell instructions somewhere)

rodrimaia (changes requested):
Code looks good, we just need to figure out how to handle the URLs

danieltodonnell (changes requested):

src/commands/cloud-login.js
```diff
@@ -67,6 +68,31 @@ async function emailStrategy(email) {
     const secret = JSON.parse(res.body).secret
     return saveEndpointOrError(newEndpoint, alias, secret)
   })
+  .catch(async function (error) {
+    // Check if call failed due to missing OTP code
+    if (JSON.parse(error.error).code === OTP_REQUIRED) {
```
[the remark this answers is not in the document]
ryancharris: @danieltodonnell Something like this?
```suggestion
    if (error.error && JSON.parse(error.error).code === OTP_REQUIRED) {
```

src/commands/cloud-login.js
```diff
@@ -4,17 +4,21 @@ const FaunaCommand = require('../lib/fauna-command.js')
 const os = require('os')
 const fetch = require('request-promise')
 const url = require('url')
+require('dotenv').config()
 
-const SHELL_LOGIN_URL = 'https://app.fauna.com/shell_login'
+const SHELL_LOGIN_URL = process.env.FAUNA_SHELL_LOGIN_URL
```
ryancharris, line 9: @BrunoQuaresma @rodrimaia How's this?

rodrimaia: @fauna-ryanharris and the default option?

rodrimaia: sorry I missed the next line haha.

rodrimaia (approved):

src/commands/cloud-login.js
```diff
@@ -4,17 +4,21 @@ const FaunaCommand = require('../lib/fauna-command.js')
...
+  ? process.env.FAUNA_SHELL_LOGIN_URL
+  : 'https://auth-console.fauna-preview.com/login';
```
ryancharris, line 11: We need to update this URL to point to the new auth-service endpoint once it's deployed 👍

BrunoQuaresma (approved):
lgtm

danieltodonnell (approved):
Code LGTM

PR #93, opened 2021-05-19
chore(release): bump to 0.12.4

parkhomenko:
### Notes
Bumped to a version 0.12.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "fauna-shell",
   "description": "faunadb shell",
-  "version": "0.12.3",
+  "version": "0.12.4",
   "author": "Fauna",
   "bin": {
-    "fauna": "bin/run"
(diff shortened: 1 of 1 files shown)


parkhomenko:
@faunaee There is no changelog in this repo, can we go without it? It would be tough to create it from scratch now. What do you think about it?

faunaee:
We're still in the 0.x series of releases for this product, so I think we can do without release notes for now. We should make it a goal to add release notes for a future release.

Other than the changes in the PR, what else is included in 0.12.4?

parkhomenko:
Here are the changes:
- shows error if user's secret is invalid
- fixed shell error (https://faunadb.atlassian.net/browse/OSS-464)
- fixed linter errors

faunaee (approved):
LGTM 👍

PR #103, opened 2021-07-05
OSS-803: wrap query if it's object

fireridlle:
### Notes
[Syntax for objects does not work as expecte](https://faunadb.atlassian.net/browse/OSS-803)

When query is not a valid JS code (like: `{ name: 'Hen Wen', age: Add(100, 10) }`), but valid FQL, shell throw an error

### DIFF
--- a/test/commands/shell.test.js
+++ b/test/commands/shell.test.js
@@ -97,7 +97,7 @@ describe('shell', () => {
     )
   })
 
-  it('run not valid JS', async () => {
+  it('run value FQL which is not valid JS', async () => {
     await new Promise((resolve, reject) => {
       shell.repl.eval(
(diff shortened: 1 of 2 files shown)


test/commands/shell.test.js
```diff
@@ -97,6 +97,23 @@ describe('shell', () => {
     )
   })
 
+  it('run not valid JS', async () => {
```
faunaee, line 100:
```suggestion
  it('run value FQL which is not valid JS', async () => {
```

faunaee (approved):
LGTM, although I made one suggestion that should imply the correct test context.

faunaee:
Note: I changed the PR's title so that Jira can auto-associate the PR with the issue id. Wrapping the Jira issue number in brackets defeats the association.

PR #114, opened 2021-08-09
chore: Bumped to 0.12.6

parkhomenko:
### Notes
Bumped to the 0.12.6 version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "fauna-shell",
   "description": "faunadb shell",
-  "version": "0.12.5",
+  "version": "0.12.6",
   "author": "Fauna",
   "bin": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "fauna-shell",
   "description": "faunadb shell",
-  "version": "0.12.5",
+  "version": "0.12.6",
```
faunaee, line 4: The most recent release was 0.12.4: https://www.npmjs.com/package/fauna-shell

We should probably use 0.12.5 as the next release. If we go with 0.12.6, some users will wonder what happened to 0.12.5.

parkhomenko: sure, I'll close this PR then and release the latest version that is already in main branch

PR #118, opened 2021-08-18
fix(oss-808): error message for eval cmd

fireridlle:
### Notes
[The eval command produces different output for queries with Abort than the shell command](https://faunadb.atlassian.net/browse/OSS-808)


faunaee (approved):
LGTM! 👍

PR #119, opened 2021-08-23
fix(oss-847): wrap json

fireridlle:
### Notes
[OSS-847 eval command produces different results than interactive query](https://faunadb.atlassian.net/browse/OSS-847)

When I run `fauna-shell` interactively, this query provides the expected result:

```
> { terms: [ { binding: 'binding1' }, { field: ['data', 'field'] } ] }
{
  terms: [ { binding: 'binding1' }, { field: [ 'data', 'field' ] } ]
}
```
However, when I run fauna eval --file=test.shell (where the file test.shell contains the query above, the result is:

```
[ { binding: 'binding1' }, { field: [ 'data', 'field' ] } ]
```
Note that the outer object and field name are missing.

### DIFF
--- a/.github/workflows/close-jira-ticket.yml
+++ b/.github/workflows/close-jira-ticket.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Login
-        uses: atlassian/gajira-login@main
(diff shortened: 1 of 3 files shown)


parkhomenko (approved):

faunaee (approved):
LGTM! 👍

PR #120, opened 2021-08-27
WIP: feature: poc data import

fireridlle:
Import data to Fauna

USAGE
  $ fauna import SOURCE

ARGUMENTS
  SOURCE  Path to .csv/.json file either to dir with .csv/.json files

OPTIONS
  --collection=collection  Collection name. By default filename if --source is file, otherwise omitted
  --db=db                  Child database name
  --domain=domain          FaunaDB server domain
  --endpoint=endpoint      FaunaDB server endpoint
  --port=port              Connection port
  --scheme=https|http      Connection scheme
  --secret=secret          FaunaDB secret key
  --timeout=timeout        Connection timeout in milliseconds

EXAMPLES
  $ fauna import --db=sampleDB ./samplefile.csv
  $ fauna import --db=sampleDB --collection=Samplecollection ./samplefile.csv
  $ fauna import --db=sampleDB ./dump


src/commands/import.js
```diff
@@ -0,0 +1,187 @@
...
+              q.Var('isCollectionExists'),
+              '',
+              q.Let(
+                {
+                  collection: q.CreateCollection({ name: collection }),
+                },
+                q.CreateIndex({
+                  source: q.Select(['ref'], q.Var('collection')),
+                  name: `all_${collection}`,
+                })
```
erickpintor, line 113: What's the reason for this index? It sounds like you could  just use the `Documents(..)` instead.

src/commands/import.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  async dataImport({ source, collection }) {
+    const { ref, isCollectionExists } = await this.upsertCollection({
+      collection,
+    })
+
+    if (isCollectionExists) {
+      await this.clearCollectionDocuments(ref)
+    }
```
erickpintor, line 57: I don't think we should clear the collection automatically. An unaware use might try to import _more_ data to a collection and ended deleting existing documents.

fireridlle:
close because of https://github.com/fauna/fauna-shell/pull/142

PR #160, opened 2022-03-17
Backoff & Retry

henryfauna:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-2042)

Add exponential backoff and retry for requests that return a `409` (Contention) or `429` (TooManyRequests).

Tests added assert the expected number of requests get made for each status code of interest.


src/lib/fauna-import-writer.js
```diff
@@ -35,19 +37,58 @@ function getFaunaImportWriter(
...
+        break
+      case 410:
+        // account disabled - exit
+        logger('Account disabled')
+        shouldRetry = false
+        break
+      case 413:
+        // request too large - reduce request size + backoff and retry
+        shouldRetry = true
+        break
```
cleve-fauna, line 56: this should never happen / is not retryable.

cleve-fauna:
include notes about what you did here and how you tested it in the description.

src/lib/fauna-import-writer.js
```diff
@@ -44,20 +46,47 @@ function getFaunaImportWriter(
...
+            })
+          )
         )
       )
-    )
+    if (allowRetries)
+      return backOff(() => write(batch), {
+        jitter: 'full',
+        retry: retryHandler,
+        numOfAttempts: 10,
```
cleve-fauna, line 87: this is a lot of retries.

Typically you'll see 3

src/lib/fauna-import-writer.js
```diff
@@ -44,20 +46,47 @@ function getFaunaImportWriter(
...
+                (memo, next) => ({ ...memo, [next.trim()]: data[next] }),
+                {}
+              ),
+            })
+          )
         )
       )
-    )
+    if (allowRetries)
+      return backOff(() => write(batch), {
```
cleve-fauna, line 84: what is the backoff algorithm?

cleve-fauna: i think you need to be more particular about configuring the backoff algorithm:

https://github.com/coveooss/exponential-backoff#ibackoffoptions

cleve-fauna: typically you'll see 500 ms - 1s for an initial backoff (the initial delay); and usage of 2 as an expontent for growing it (the time multiple).

src/commands/import.js
```diff
@@ -145,7 +145,8 @@ class ImportCommand extends FaunaCommand {
           collection,
           path,
           Boolean(this.flags['dry-run']),
-          this.warn
+          this.warn,
+          Boolean(this.flags['allow-retries'])
```
cleve-fauna, line 149: IMO we don't need a flag for this - we're retrying things that we know didn't persist and are retryable. We should just do this for the user without needing configuration.

src/lib/fauna-import-writer.js
```diff
@@ -25,6 +26,7 @@ function getFaunaImportWriter(
   inputFile,
   isDryRun = false,
   logger = console.log,
+  allowRetries = false,
```
cleve-fauna, line 29: is the reason this is a parameter to keep existing tests functional?

IMO we'd be better off just always retrying and updating our tests to account for this.

package-lock.json
```diff
@@ -1,10678 +1,8 @@
 {
   "name": "fauna-shell",
   "version": "0.14.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
cleve-fauna, line 4: this is changing the format of the package-lock

cleve-fauna (approved):

PR #163, opened 2022-03-23
Implement dynamic rate limiting

elersong:
### Notes
[Jira Ticket]()

### Screenshots

### DIFF
--- a/src/lib/fauna-command.js
+++ b/src/lib/fauna-command.js
@@ -85,6 +85,11 @@ class FaunaCommand extends Command {
     return errorOut(err, 1)
   }
 
+  getQueryStats(res) {
+    if (!res || !res.responseHeaders) return
+    console.log(res.responseHeaders)
+  }
+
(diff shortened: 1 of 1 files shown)


cleve-fauna:
Add some notes to your PRs.

### Notes

What this is for.

How you did it.

Any caveats, things out of scope, etc.

### Testing

How you tested

src/lib/fauna-command.js
```diff
@@ -85,7 +85,12 @@ class FaunaCommand extends Command {
     return errorOut(err, 1)
   }
 
-  async getClient({ dbScope, role } = {}) {
+  getQueryStats(res) {
+    if (!res || !res.responseHeaders) return
+    console.log(res.responseHeaders)
```
cleve-fauna, line 90: is this intended just to be a POC?

How can we get this out to the user of the client so we can consume the stats?

henryfauna:
Probably a good idea to add a test verifying you get the headers you expect as well

PR #166, opened 2022-03-24
[FE-2101] Switch rate limiting library

henryfauna:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-2101)

Switch rate limiting library from [limiter](https://www.npmjs.com/package/limiter) to [rate-limiter-flexible](https://www.npmjs.com/package/rate-limiter-flexible).

### DIFF
--- a/src/lib/fauna-import-writer.js
+++ b/src/lib/fauna-import-writer.js
@@ -46,12 +46,11 @@ function getFaunaImportWriter(
   })
 
   const applyRateLimitPenalty = () => {
-    const bytesPerSecondFloor = 50000
-    if (rateLimiter.points > bytesPerSecondFloor) rateLimiter.points /= 2
+    rateLimiter.points /= 2
   }
 
   const bumpRateLimit = () => {
-    const increment = 10000
+    const increment = bytesPerSecondLimit / 100
     if (rateLimiter.points < bytesPerSecondLimit - increment) {
       rateLimiter.points += increment
     }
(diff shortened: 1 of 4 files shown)


src/lib/fauna-import-writer.js
```diff
@@ -37,10 +37,22 @@ function getFaunaImportWriter(
...
+
   const waitForRateLimitTokens = (tokens, rateLimiter) => {
-    while (!rateLimiter.tryRemoveTokens(tokens)) {
+    const tryToConsumeTokens = () => {
+      return rateLimiter
+        .consume('', tokens)
+        .then(() => true)
+        .catch(() => false)
+    }
+    while (!tryToConsumeTokens()) {
```
cleve-fauna, line 55: `const tryToConsumeTokens` is a promise.

Evaluating its truthiness will always evaluate to true.

You need to use async / await to acheive your goal here of waiting until there's enough tokens.

cleve-fauna: A test proving the rate limiter's behavior would have caught this. Its probably worth adding one.

cleve-fauna: a trick i've used to test rate limiters before is something like:

```
start timer
set rate limit to like 1 / s
issue ~3 requests
confirm timer >= 3
```

usually pretty easy to code up.

mwilde345 (approved):

cleve-fauna (approved):
This is a good start.

I think we should go ahead and do another commit / PR where we consume the remainder of the work in the case the `Math.min(bytesPerSecondLimit, dataSize)` uses the bytesPersSecondLimit (more specifically dataSize > bytePerSecondLmit).

Otherwise we leave ourselves open to letting something we shouldn't through.

PR #208, opened 2023-07-19
ENG-5277: Implement fauna schema:ls

wdberkeley:
This adds CLI support for listing schema files in a scope.

I would have liked the command to be `fauna schema ls`, but we are using a pre-v2 version of oclif, and it only supports : as the separator. Maybe we should upgrade?

### DIFF
--- a/src/commands/schema/ls.js
+++ b/src/commands/schema/ls.js
@@ -1,0 +1,44 @@
+const FaunaCommand = require('../../lib/fauna-command.js')
+const fetch = require('node-fetch')
+const { errorOut } = require('../../lib/misc.js')
+
+class ListSchemaCommand extends FaunaCommand {
+  async run() {
+    const log = this.log
+    const {
+      connectionOptions: { domain, port, scheme, secret },
+    } = await this.getClient()
+
+    return fetch(`${scheme}://${domain}:${port}/schema/1/files`, {
+      method: 'GET',
+      headers: { AUTHORIZATION: `Bearer ${secret}` },
+    })
+      .then(function (res) {
+        return res.json()
+      })
+      .then(function (json) {
+        if (json.files.length > 0) {
+          log('Schema files')
+          json.files.forEach(function (file) {
+            log(file.filename)
+          })
+        } else {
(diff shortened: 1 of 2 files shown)


macmv:
we definitely should upgrade, thats just a bit of a refactor. I'm happy to help upgrade if you think its worth the time

src/commands/schema/ls.js
```diff
@@ -0,0 +1,44 @@
...
+          json.files.forEach(function (file) {
+            log(file.filename)
+          })
+        } else {
+          log('No schema files')
+        }
+      })
+      .catch(function (err) {
+        errorOut(err)
+      })
```
macmv, line 31: style: js has `await` for this:
```suggestion
    try {
      const res = await fetch(`${scheme}://${domain}:${port}/schema/1/files`, {
        method: 'GET',
        headers: { AUTHORIZATION: `Bearer ${secret}` },
      });
      const json = await res.json();
      if (json.files.length > 0) {
        log('Schema files')
        json.files.forEach(function (file) {
          log(file.filename)
        })
      } else {
        log('No schema files')
      }
    } catch (e) {
      errorOut(e);
    }
```

src/commands/schema/ls.js
```diff
@@ -0,0 +1,44 @@
...
+    return fetch(`${scheme}://${domain}:${port}/schema/1/files`, {
+      method: 'GET',
+      headers: { AUTHORIZATION: `Bearer ${secret}` },
+    })
+      .then(function (res) {
+        return res.json()
+      })
+      .then(function (json) {
+        if (json.files.length > 0) {
+          log('Schema files')
```
macmv, line 21: can you add a newline here? It might be nice to have a separator between this header and the actual files

macmv (commented):
looking good, just some minor feedback. also we need to add you to this workspace, I think you're pring from a fork right now.

erickpintor (approved):
Besides styling, feature wise looks good to me.

wdberkeley:
Gonna redo this as a PR from a branch in this repo...

PR #209, opened 2023-07-19
Add eval v10

macmv:
Ticket: FE-3712

Adds `--version 4/10` to `fauna eval`. This should not break existing usage of the command.

When `--version 10` is set, the following behavior is changed:
- the `format` now defaults to "shell" when in a tty, and "json" when piping to another command
- the query string is sent to the v10 endpoint instead of being evaled
  - the "shell" format will set the query format to "decorated"
  - the "json" format will set the query format to "simple"
- a new `json-tagged` format has been added. this does the same thing as `json` in v4, but returns the tagged value in v10 (instead of the simple value).

### DIFF
--- a/src/commands/eval.js
+++ b/src/commands/eval.js
@@ -6,7 +6,7 @@ const faunadb = require("faunadb");
 const FaunaCommand = require("../lib/fauna-command.js");
 const { readFile, runQueries, writeFile } = require("../lib/misc.js");
 
-const EVAL_OUTPUT_FORMATS = ["json", "shell"];
+const EVAL_OUTPUT_FORMATS = ["json", "json-tagged", "shell"];
 
 /**
  * Write json encoded output
@@ -14,12 +14,13 @@ const EVAL_OUTPUT_FORMATS = ["json", "shell"];
  * @param {String} file Target filename
  * @param {Any}    data Data to encode
  */
-function writeFormattedJson(file, data) {
+async function writeFormattedJson(file, data) {
   let str = JSON.stringify(data);
   if (file === null) {
-    return Promise.resolve(console.log(str));
+    console.log(str);
+  } else {
+    await writeFile(file, str);
   }
-  return writeFile(file, str);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/lib/fauna-client.js
```diff
@@ -0,0 +1,71 @@
...
+        req = this.session
+          .request(httpRequestHeaders)
+          .setEncoding("utf8")
+          .on("error", (error) => rejectPromise(error))
+          .on("response", onResponse);
+
+        req.write(JSON.stringify({ query }), "utf8");
+
+        // req.setTimeout must be called before req.end()
+        req.setTimeout(5000, () => {
```
ptpaterson, line 57: We should set the client timeout to the `--timeout` parameter + 5000 ms. If no `--timeout` is provided, then maybe 10000 ms is okay.

According to docs, if no `--timeout` is provided, then we should let the query run. Default query timeout for v10 is 5 seconds, so client timeout of 5 seconds in that case is too short.

https://docs.fauna.com/fauna/current/build/tools/shell/commands/eval#options

src/commands/eval.js
```diff
@@ -110,19 +82,131 @@ class EvalCommand extends FaunaCommand {
...
+      return this.error("Unsupported output format");
+    }
+  }
+
+  // Flags should be
+  // {
+  //   version: "4" | "10";
+  //   format: "json" | "json-tagged" | "shell";
+  //   typecheck?: boolean;
+  // }
```
ptpaterson, line 154: If we need clarification with comments, can you use JS docs style comments?

ptpaterson (changes requested):

src/commands/eval.js
```diff
@@ -146,12 +146,14 @@ class EvalCommand extends FaunaCommand {
     }
   }
 
-  // Flags should be
-  // {
-  //   version: "4" | "10";
-  //   format: "json" | "json-tagged" | "shell";
-  //   typecheck?: boolean;
-  // }
+  /**
```
ptpaterson, line 149: 😂 My bad, I meant to use JS docs like with `@param` lines. I can take a stab at it since this is my nitpick.

macmv: I dislike `@param` lines because they make it difficult to read. This is just my dislike of javadoc syntax, but I'd prefer to just use words in comments is all.

If you feel strongly about it feel free to go add the `@param` bit, I just think it makes things harder to read.

macmv: makes sense I guess. I don't use those docs lol, I just rely on typescript signatures.

macmv:
lol hi `@param`

src/commands/eval.js
ptpaterson: Since this is JS and not TS, there is no TS signature.  The JS docs can actually give you a TS signature.

current signature preview
image

updated signature preview
image

ptpaterson (approved):

PR #210, opened 2023-07-20
ENG-5277: Implement fauna schema:ls

wdberkeley:
This adds CLI support for listing schema files in a scope.

I would have liked the command to be `fauna schema ls`, but we are using a pre-v2 version of oclif, and it only supports : as the separator. I'm planning to submit an upgrade + fix PR after this series of command implementations.


erickpintor:
>  I'm planning to submit an upgrade + fix PR after this series of command implementations.

Could you log a ticket for it? I think this is an important UX improvement.

erickpintor (approved):
:sheep: it.

PR #211, opened 2023-07-20
ENG-5273: Implement fauna schema:pull

wdberkeley:
This implements fauna schema pull, but it prints the contents of the file to stdout instead of saving them in the CWD. I think this is better for a single schema file because stdout can be redirected to send the file's contents elsewhere.

We can change it later when we support multiple files in a tree.

### DIFF
--- a/src/commands/schema/pull.js
+++ b/src/commands/schema/pull.js
@@ -1,0 +1,43 @@
+const FaunaCommand = require('../../lib/fauna-command.js')
+const fetch = require('node-fetch')
+const { errorOut } = require('../../lib/misc.js')
+
+class PullSchemaCommand extends FaunaCommand {
+  async run() {
+    const log = this.log
+    const filename = this.args.filename
+    const {
+      connectionOptions: { domain, port, scheme, secret },
+    } = await this.getClient()
+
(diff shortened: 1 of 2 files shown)


erickpintor (approved):
I think this diverges from the tool intent a bit as it's supposed to be a file manager like the `aws s3` command line tool. It's fine for a first demo, if you want to mere it as is, but I'd say we follow up quickly after with file a management ticket.

erickpintor:
> I think this is better for a single schema file because stdout can be redirected to send the file's contents elsewhere.

BTW: I agree with the value of printing out the file without storing, I'd just put that in a separate command. Like `fauna shema cat main.fsl`.

macmv (approved):
I agree with the comments erick left. Seems good for now, but we should probably improve this soon.

wdberkeley:
> > I think this is better for a single schema file because stdout can be redirected to send the file's contents elsewhere.
> 
> BTW: I agree with the value of printing out the file without storing it. I'd just put that in a separate command, like `fauna shema cat main.fsl`.

Yeah, and what I'm thinking is we just rename this one and make a copy of its code as a starting point for the more sophisticated fs stuff. This ticks all the boxes for the steel thread, and it more convenient to use besides.

PR #212, opened 2023-07-20
ENG-5274: Implement fauna schema:pull

wdberkeley:


main.fsl
```diff
@@ -0,0 +1 @@
+collection Foo {}
```
macmv, line 1: maybe put this in a test directory? adding `main.fsl` to the project root seems a bit strange

package-lock.json
```diff
@@ -11,6 +11,7 @@
       "dependencies": {
         "@oclif/command": "^1.5.19",
         "@oclif/config": "^1.13.3",
+        "@oclif/core": "^2.9.4",
```
macmv, line 14: this will cause many problems. I'm working on porting everything to the new version, I think it'll be best to just stick with `@oclif/command` for now. `@oclif/core` is a rewrite essentially, so I think it'll break things to have both.

macmv (commented):
looks good, but don't add `@oclif/core`. I believe that will duplicate the whole oclif dependency.

erickpintor (commented):
Functionality wise, looks good. How can we address the dependency concern, though?

erickpintor (approved):
lgtm

macmv (approved):

erickpintor (approved):
re- 🐑 it.

PR #228, opened 2023-08-28
ENG-5275: Implement fauna schema {push,pull}

wdberkeley:
This implements fauna schema push and fauna schema pull with the following semantics:

fauna schema push requires a directory set with --dir. It collects all .fsl files in the directory and submits them to the validation endpoint, displaying the diff to the user. If the user accepts the change, the files are pushed at the schema version returned by the validate call. Optionally, users may skip version checks and validation with --force, or specify their own schema version to check against with --version.

fauna schema pull requires a directory set with --dir. By default, it deletes all .fsl files (not directories ending in .fsl) and downloads the schema as .fsl files into the directory. Users may keep .fsl files that are not part of the download by setting --retain.

I also restored the fauna schema ls and fauna schema cat commands.

While there is an automated test for parts of the code, I also manually tested the file handling behavior with a local Fauna instance.

### DIFF
--- a/src/commands/schema/cat.js
+++ b/src/commands/schema/cat.js
@@ -1,0 +1,41 @@
+const SchemaCommand = require("./schema-command.js");
+const { Args } = require("@oclif/core");
+const fetch = require("node-fetch");
+
+class CatSchemaCommand extends SchemaCommand {
+  async run() {
+    const filename = this.args.filename;
+    const { urlbase, secret } = await this.fetchsetup();
+
+    try {
+      const res = await fetch(`${urlbase}/schema/1/files/${filename}`, {
+        method: "GET",
+        headers: { AUTHORIZATION: `Bearer ${secret}` },
+      });
+      const json = await res.json();
+      if (json.error) {
+        this.error(json.error.message);
+      } else {
+        this.log(json.content);
+      }
+    } catch (err) {
+      this.error(err);
+    }
+  }
+}
+
+CatSchemaCommand.description = "Display the contents of a schema file";
+
+CatSchemaCommand.examples = ["$ fauna schema cat main.fsl"];
+
+CatSchemaCommand.args = {
+  filename: Args.string({
+    required: true,
+    description: "name of schema file",
+  }),
+};
+
+CatSchemaCommand.flags = {
+  ...SchemaCommand.flags,
+};
+module.exports = CatSchemaCommand;
--- a/src/commands/schema/ls.js
+++ b/src/commands/schema/ls.js
@@ -1,0 +1,39 @@
+const SchemaCommand = require("./schema-command.js");
+const fetch = require("node-fetch");
+
+class ListSchemaCommand extends SchemaCommand {
+  async run() {
+    const { urlbase, secret } = await this.fetchsetup();
+
+    try {
+      const res = await fetch(`${urlbase}/schema/1/files`, {
+        method: "GET",
+        headers: { Authorization: `Bearer ${secret}` },
+      });
+      const json = await res.json();
+      if (json.error) {
+        this.error(json.error.message);
+      } else if (json.files.length > 0) {
+        this.log("Schema files:\n");
+        json.files.forEach((file) => {
+          this.log(file.filename);
+        });
+      } else {
+        this.log("No schema files");
+      }
+    } catch (err) {
+      this.error(err);
+    }
+  }
+}
+
+ListSchemaCommand.description = "List database schema files";
+
+ListSchemaCommand.examples = ["$ fauna schema ls"];
+
+ListSchemaCommand.args = [];
+
+ListSchemaCommand.flags = {
+  ...SchemaCommand.flags,
+};
+module.exports = ListSchemaCommand;
--- a/src/commands/schema/pull.js
+++ b/src/commands/schema/pull.js
@@ -1,0 +1,79 @@
+const SchemaCommand = require("./schema-command.js");
+const fetch = require("node-fetch");
+const fs = require("fs");
+const path = require("path");
+const { Flags } = require("@oclif/core");
+
+class PullSchemaCommand extends SchemaCommand {
+  async run() {
+    const { urlbase, secret } = await this.fetchsetup();
+
+    const dir = this.flags.dir;
+
+    try {
+      const filesres = await fetch(`${urlbase}/schema/1/files`, {
+        method: "GET",
+        headers: { AUTHORIZATION: `Bearer ${secret}` },
+      });
+      const filesjson = await filesres.json();
+      if (filesjson.error) {
+        this.error(filesjson.error.message);
+      }
+      // Sort for consistent order. It's nice for tests.
+      const filenames = filesjson.files
+        .map((file) => file.filename)
+        .filter((name) => name.endsWith(".fsl"))
+        .sort();
+
+      // All of the below could be parallelized if necessary.
+      if (!this.flags.retain) {
+        // Delete all .fsl files (not directories).
+        for (const deleteme of fs.readdirSync(dir)) {
+          const fp = path.join(dir, deleteme);
+          if (deleteme.endsWith(".fsl") && !fs.statSync(fp).isDirectory()) {
+            fs.unlinkSync(fp);
+          }
+        }
+      }
+
+      for (const filename of filenames) {
+        const fileres = await fetch(`${urlbase}/schema/1/files/${filename}`, {
+          method: "GET",
+          headers: { AUTHORIZATION: `Bearer ${secret}` },
+        });
+        const filejson = await fileres.json();
+        if (filejson.error) {
+          this.error(filejson.error.message);
+        }
+        const fp = path.join(dir, filename);
+        fs.mkdirSync(path.dirname(fp), { recursive: true });
+        fs.writeFileSync(fp, filejson.content);
+      }
+    } catch (err) {
+      this.error(err);
+    }
+  }
+}
+
+PullSchemaCommand.description =
+  "Pull a database schema's .fsl files into a directory";
+
+PullSchemaCommand.examples = [
+  "$ fauna schema pull --dir schemas/myschema --retain",
+];
+
+PullSchemaCommand.flags = {
+  ...SchemaCommand.flags,
+  retain: Flags.boolean({
+    description:
+      "Retain .fsl files in the target directory that are not part of the database schema",
+    default: false,
+  }),
+  // NB: Required as a flag because it will become optional eventually,
+  //     once project configuration is implemented.
+  dir: Flags.string({
+    required: true,
+    description: "The target directory",
+  }),
+};
+module.exports = PullSchemaCommand;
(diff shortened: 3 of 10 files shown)


src/commands/schema/cat.js
```diff
@@ -0,0 +1,46 @@
+const FaunaCommand = require("../../lib/fauna-command.js");
+const { Args } = require("@oclif/core");
+const fetch = require("node-fetch");
+
+class CatSchemaCommand extends FaunaCommand {
```
macmv, line 5: i'm not sure if we wanted to keep this command. its kinda useful, but doesn't fit with the project management stuff all that well

src/commands/schema/cat.js
```diff
@@ -0,0 +1,46 @@
...
+  async run() {
+    const filename = this.args.filename;
+    const {
+      connectionOptions: { domain, port, scheme, secret },
+    } = await this.getClient();
+
+    try {
+      const res = await fetch(
+        `${scheme}://${domain}:${port}/schema/1/files/${filename}`,
```
macmv, line 14: does this work if `domain` and `port` aren't set in the `~/.fauna-shell` file? I ran into a bug related to that with `eval --version 10`. It might be better to move this into a more general `endpoint` function on the client.

src/commands/schema/ls.js
```diff
@@ -0,0 +1,41 @@
+const FaunaCommand = require("../../lib/fauna-command.js");
+const fetch = require("node-fetch");
+
+class ListSchemaCommand extends FaunaCommand {
```
macmv, line 4: minor: I'm assuming this lists all files, even one in nested directories? if so, we might want to name it something other than `ls`, as that isn't how `ls` behaves in the shell.

src/commands/schema/pull.js
```diff
@@ -0,0 +1,85 @@
...
+        .filter((name) => name.endsWith(".fsl"))
+        .sort();
+
+      // All of the below could be parallelized if necessary.
+      if (this.flags.delete) {
+        // Delete all .fsl files (not directories).
+        for (const deleteme of fs.readdirSync(dir)) {
+          const fp = path.join(dir, deleteme);
+          if (deleteme.endsWith(".fsl") && !fs.statSync(fp).isDirectory()) {
+            fs.unlinkSync(fp);
```
macmv, line 39: lol this is a terrible function name, thank you javascript.

will this explode if you can't write the file? we might want some error handling to show the path and such.

also note that `statSync` can fail if you can't read the file, and I think that needs to be handled as well

src/commands/schema/pull.js
```diff
@@ -0,0 +1,85 @@
...
+          {
+            method: "GET",
+            headers: { AUTHORIZATION: `Bearer ${secret}` },
+          }
+        );
+        const filejson = await fileres.json();
+        if (filejson.error) {
+          this.error(filejson.error.message);
+        }
+        fs.writeFileSync(path.join(dir, filename), filejson.content);
```
macmv, line 56: I think this needs a `fs.mkdirSync(path.parent(path.join(dir, filename)), { recursive: true })` to make the parent directories for these files

src/commands/schema/pull.js
```diff
@@ -0,0 +1,85 @@
...
+        this.error(filesjson.error.message);
+      }
+      // Sort for consistent order. It's nice for tests.
+      const filenames = filesjson.files
+        .map((file) => file.filename)
+        .filter((name) => name.endsWith(".fsl"))
+        .sort();
+
+      // All of the below could be parallelized if necessary.
+      if (this.flags.delete) {
```
macmv, line 34: shouldn't this be
```suggestion
      if (!this.flags.retain) {
```

src/commands/schema/push.js
```diff
@@ -0,0 +1,127 @@
...
+      connectionOptions: { domain, port, scheme, secret },
+    } = await this.getClient();
+    const dir = this.flags.dir;
+    const filenames = fs.readdirSync(dir);
+    var totalsize = 0;
+    const files = filenames
+      .filter(
+        (filename) =>
+          filename.endsWith(".fsl") &&
+          !fs.statSync(path.join(dir, filename)).isDirectory()
```
macmv, line 23: files can be many things. this should probably be `isFile`, so that symlinks and other weird things aren't uploaded.
```suggestion
          fs.statSync(path.join(dir, filename)).isFile()
```

src/commands/schema/push.js
```diff
@@ -0,0 +1,127 @@
...
+            body: body(),
+          }
+        );
+        const json = await res.json();
+        if (json.error) {
+          this.error(json.error.message);
+        }
+      } else {
+        // Confirm diff, then push it.
+        var vurl = `${scheme}://${domain}:${port}/schema/1/validate?force=true`;
```
macmv, line 66: remind me, what does `?force` do on `/validate`?

also this should use `let` not `var`. var declares a function-global variable in JS, whereas `let` has the same scoping mechanics as `const`

src/commands/schema/push.js
```diff
@@ -0,0 +1,127 @@
...
+        });
+        const json = await res.json();
+        if (json.error) {
+          this.error(json.error.message);
+          return;
+        }
+        this.log(`Proposed diff:\n`);
+        this.log(json.diff);
+        if (await ux.confirm("Accept and push the changes?")) {
+          var purl = `${scheme}://${domain}:${port}/schema/1/update?version=${json.version}`;
```
macmv, line 83: same here, `let` not `var`.

src/commands/schema/cat.js
macmv: nvm we're keeping them

src/commands/schema/cat.js
freels: I think it would be prudent to pull the duplicated URL generation logic out to a helper (at least the part which handles scheme, domain, and port.)

It also doesn't look like these commands use the endpoint configuration at all?

test/commands/schema.test.js
```diff
@@ -0,0 +1,132 @@
...
+        .post("/schema/1/update?version=0")
+        .reply(200, updated)
+    )
+    .stdout()
+    .command(withOpts(["schema push", "--dir=test/testdata"]))
+    .it("runs schema push", (ctx) => {
+      expect(ctx.stdout).to.contain(`${diff.diff}`);
+    });
+});
+
```
freels, line 94: would be good to test both this and push with some nested directory structure.

The --retain behavior needs testing as well.

freels (changes requested):

src/commands/schema/cat.js
freels: followup: I see it's getting these from the client. That was not obvious (rusty JS here...). Seems like that at least needs to be explained or cleaned up, though.

wdberkeley: Sorry, is there a particular thing needed from this change?

src/commands/schema/ls.js
wdberkeley: It lists the schema files for one database and not for databases nested inside, so I think it behaves like `ls` in that way.

src/commands/schema/pull.js
wdberkeley: This is all inside the try, which will spit out the error on the console. Is that not sufficient?

src/commands/schema/pull.js
wdberkeley: Yeah, good point. The dir given might not exist yet.

wdberkeley: Actually, I take it back. The user should make the directory. They should set the permissions, etc.

wdberkeley: I could see us taking a dir and making a subdirectory of that directory for all the files, but not making the top-level directory ourselves. Some extension of the tool might do something like that.

src/commands/schema/pull.js
wdberkeley: Oops.

src/commands/schema/push.js
wdberkeley: I think if we can open it and suck bytes out of it like a file we can upload it. I wouldn't use symlinks myself but I would expect it to read file contents through a symlink.

Plus, if you try to handle every goddamn unix this-or-that then your head will explode. Well-established fact. The way to deal with people doing weird dumb stuff is to tell them not to do that.

src/commands/schema/push.js
wdberkeley: Even validate wants a version to do a version check, so `force` skips the check so we can get back a version to use for push.

src/commands/schema/cat.js
freels: I'd appreciate if fetch setup were factored out into a shared helper (maybe a shared SchemaCommand superclass?). And testing to ensure that domain and port are either always set, or that the code can handle those not being present (I haven't looked at the internals, just re-raising @macmv's concern).

test/commands/schema.test.js
wdberkeley: I added some more annoying javascript tests.

src/commands/schema/pull.js
jdcryans: Note from standup: Neil's comment above was for subdirectories which we do support.

Regarding the top level, we should do what rsync does.

src/commands/schema/pull.js
macmv: thats probably alright, but the error won't show the path of the file. It'd be nice to wrap this `if` in a try that shows logs `fp` on an error.

wdberkeley: It looks like it does show good information. Here's what it looked like when I tested push on an unknown directory:

```
$ bin/run schema push --dir=$HOME/aslkdfjgalksdfj
    Error: ENOENT: no such file or directory, scandir '/Users/wdberkeley/aslkdfjgalksdfj'
    Code: ENOENT
```

test/commands/schema.test.js
```diff
@@ -0,0 +1,166 @@
...
+const testdir = "test/testdata";
+
+const setup = () => {
+  try {
+    fs.unlinkSync(path.join(testdir, "functions.fsl"));
+  } catch (err) {
+    // 2023 technology.
```
macmv, line 101: lol

src/commands/schema/schema-command.js
```diff
@@ -0,0 +1,16 @@
+const FaunaCommand = require("../../lib/fauna-command.js");
+
+class SchemaCommand extends FaunaCommand {
+  async fetchsetup() {
+    const {
+      connectionOptions: { domain, port, scheme, secret },
+    } = await this.getClient();
+
+    return {
+      urlbase: `${scheme}://${domain}:${port ? port : "4443"}`,
```
macmv, line 10:
```suggestion
      urlbase: `${scheme ?? "https"}://${domain}${port ? `:${port}` : ""}`,
```

macmv (commented):
looks good, just one suggested change on urlbase

macmv (approved):

freels (approved):

PR #234, opened 2023-08-30
ENG-5464: Make fauna schema push recursive

wdberkeley:
This got missed before. `fauna schema push` will now recur into all subdirectories and push all fsl files with a name equal to the relative path to the .fsl file. So, `fauna schema push --dir=/foo/bar` will push `/foo/bar/a/b/functions.fsl` as an FSL file named "a/b/functions.fsl".

It isn't amenable to the unit tests but I tested manually that the recursive traversal works to multiple levels of depth.


macmv (approved):
looks good. im assuming `pull` changes for files in directories are coming soon?

PR #238, opened 2023-08-31
ENG-5479: Default to "no" when confirming changes

wdberkeley:
I had to write more code than you would think because ux.confirm doesn't support defaults.

### DIFF
--- a/src/commands/schema/push.js
+++ b/src/commands/schema/push.js
@@ -9,6 +9,20 @@ const FILE_LIMIT = 256;
 const FILESIZE_LIMIT_BYTES = 32 * 1024 * 1024;
 
 class PushSchemaCommand extends SchemaCommand {
+  async confirm() {
+    const resp = await ux.prompt("Accept and push the changes?", {
+      default: "no",
+    });
+    if (["yes", "y"].includes(resp.toLowerCase())) {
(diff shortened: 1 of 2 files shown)


src/commands/schema/push.js
```diff
@@ -9,6 +9,20 @@ const FILE_LIMIT = 256;
 const FILESIZE_LIMIT_BYTES = 32 * 1024 * 1024;
 
 class PushSchemaCommand extends SchemaCommand {
+  async confirm() {
+    const resp = await ux.prompt("Accept and push the changes?", {
```
macmv, line 13:
```suggestion
    const resp = await ux.prompt("Accept and push the changes? (y/N)", {
```

wdberkeley: Because of the default, the prompt shows up as

Accept and push the changes? [no]:

so putting (y/N) in there is redundant.

macmv: putting `[no]` in there is more confusing though, as its not clear what the options are

wdberkeley: If our users can't figure out this is a yes or no question, we are in big trouble.

macmv (approved):

PR #246, opened 2023-09-13
Add typescript

macmv:
Ticket(s): FE-###

Adds the typescript compiler, and switches `list-databases` to be implemented in typescript, to make sure it works.


src/commands/list-databases.ts
```diff
@@ -0,0 +1,51 @@
...
+              });
+            } else {
+              this.log("No databases created");
+            }
+          })
+          .catch((err) => {
+            this.error(err.message);
+          });
+      },
+      undefined as any, // is dbScope, an optional param
```
wdberkeley, line 35: Typescript makes it necessary to specify "optional" parameters like this?

PR #247, opened 2023-09-13
Add project config

macmv:
Ticket(s): ENG-5276

Reads `.fauna-project` files, and adds a new config system that should allow us to access the config more easily.

One side affect of this PR is that all secrets must be admin secrets. This was not required before, but now it is, for consistency with schema commands (which will always require admin secrets).

### DIFF
--- a/src/lib/config/project-config.ts
+++ b/src/lib/config/project-config.ts
@@ -1,0 +1,51 @@
+import { RootConfig, Config, InvalidConfigError } from ".";
+
+// Represents `.fauna-project`
+export class ProjectConfig {
+  defaultStack?: string;
+  stacks: { [key: string]: Stack };
+
+  constructor(config: Config) {
+    this.defaultStack = config.strOpt("default");
+    this.stacks = Object.fromEntries(
+      config.objectsIn("stack").map(([k, v]) => [k, new Stack(v)])
+    );
+
+    if (this.defaultStack === "default") {
+      throw new InvalidConfigError("Default stack cannot be named 'default'");
+    } else if (
+      this.defaultStack &&
+      this.stacks[this.defaultStack] === undefined
+    ) {
+      throw new InvalidConfigError(
+        `Default stack '${this.defaultStack}' was not found`
+      );
+    }
+  }
+
+  validate(rootConfig: RootConfig) {
+    for (const stack of Object.values(this.stacks)) {
+      if (rootConfig.endpoints[stack.endpoint] === undefined) {
+        throw new InvalidConfigError(
+          `Endpoint '${stack.endpoint}' not found in ~/.fauna-shell`
+        );
+      }
+    }
+  }
+}
+
+export class Stack {
+  /**
+   * The endpoint name to use as a base.
+   */
+  endpoint: string;
+  /**
+   * The database path to use.
+   */
+  database?: string;
+
+  constructor(config: Config) {
+    this.endpoint = config.str("endpoint");
+    this.database = config.str("database");
+  }
+}
--- a/src/lib/config/project-config.ts
+++ b/src/lib/config/project-config.ts
@@ -11,12 +11,7 @@ export class ProjectConfig {
       config.objectsIn("stack").map(([k, v]) => [k, new Stack(v)])
     );
 
-    if (this.defaultStack === "default") {
-      throw new InvalidConfigError("Default stack cannot be named 'default'");
-    } else if (
-      this.defaultStack &&
-      this.stacks[this.defaultStack] === undefined
-    ) {
+    if (this.defaultStack && this.stacks[this.defaultStack] === undefined) {
       throw new InvalidConfigError(
         `Default stack '${this.defaultStack}' was not found`
       );
--- a/src/lib/config/project-config.ts
+++ b/src/lib/config/project-config.ts
@@ -37,7 +37,7 @@ export class Stack {
   /**
    * The database path to use.
    */
-  database?: string;
+  database: string;
 
   constructor(config: Config) {
     this.endpoint = config.str("endpoint");
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,6 +24,7 @@
       }
     ]
   },
+  "ignorePatterns": ["dist/**/*"],
   "overrides": [
     {
       "files": ["*.test.js", "*.spec.js"],
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "build": "rm -rf dist && tsc -b",
     "postpack": "rm -f oclif.manifest.json",
     "prepack": "yarn build && oclif manifest",
-    "pretest": "npm run fixlint",
+    "pretest": "yarn fixlint",
     "local-test": "export $(cat .env | xargs); nyc mocha --forbid-only \"test/**/*.test.{js,ts}\"",
     "test": "export $(cat .env.test | xargs); nyc mocha --forbid-only \"test/**/*.test.{js,ts}\"",
     "lint": "eslint .",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "lint": "eslint .",
     "fixlint": "eslint . --fix",
     "version": "oclif-dev readme && git add README.md",
-    "fmt": "prettier -w ."
+    "fmt": "prettier -w src"
   },
   "husky": {
     "hooks": {
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -249,6 +249,7 @@ export class ShellConfig {
       }
 
       // override endpoint with values from flags.
+      this.endpoint.secret = secretFlag ?? this.endpoint.secret;
       this.endpoint.url = urlFlag ?? this.endpoint.url;
       this.endpoint.graphqlHost =
         this.flags.strOpt("graphqlHost") ?? this.endpoint.graphqlHost;
@@ -258,22 +259,21 @@ export class ShellConfig {
   }
 
   lookupEndpoint = (): EndpointConfig => {
-    return this.endpoint.makeScopedEndpoint(
-      this.args.scope ?? this.stack?.database,
-      this.args.role
-    );
+    let database = this.stack?.database ?? "";
+    if (this.stack !== undefined && this.args.scope !== undefined) {
+      database += "/";
+    }
+    database += this.args.scope;
+
+    return this.endpoint.makeScopedEndpoint(database, this.args.role);
   };
 }
 
 const readFileOpt = (fileName: string) => {
-  try {
+  if (fs.existsSync(fileName)) {
     return fs.readFileSync(fileName, "utf8");
-  } catch (err) {
-    if (isFileNotFound(err)) {
-      return "";
-    } else {
-      throw err;
-    }
+  } else {
+    return "";
   }
 };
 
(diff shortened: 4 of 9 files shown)


src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  constructor(config: Config) {
+    this.defaultEndpoint = config.strOpt("default");
+    this.endpoints = Object.fromEntries(
+      config
+        .entries((k) => k !== "default")
+        .map(([k, v]) => [k, new Endpoint(v)])
+    );
+
+    if (this.defaultEndpoint === "default") {
```
fauna-chase, line 103: would this be like:

```
default = default

[default]
secret = blah

```
?

Would it fail to parse in that scenario?

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+      throw new InvalidConfigError(
+        `Default endpoint '${this.defaultEndpoint}' was not found`
+      );
+    }
+  }
+}
+
+// Represents `.fauna-project`
+class LocalConfig {
+  defaultStack: string | undefined;
```
fauna-chase, line 120: is this equivalent to the above:
```
defaultStack?: string
```
?

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+    ) {
+      throw new InvalidConfigError(
+        `Default endpoint '${this.defaultEndpoint}' was not found`
+      );
+    }
+  }
+}
+
+// Represents `.fauna-project`
+class LocalConfig {
```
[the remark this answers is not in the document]
fauna-chase: could we also name this class ProjectConfig?

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  constructor(config: Config) {
+    this.secret = config.str("secret");
+    this.url = getURLFromConfig(config);
+
+    this.graphqlHost = config.strOpt("graphqlHost");
+    this.graphqlPort = config.numberOpt("graphqlPort");
+  }
+
+  makeEndpoint(
```
fauna-chase, line 167: wondering if we could call this something like:
`withScopeOrRole` and then just return a new Endpoint object hat is updated accordingly.

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+   */
+  database: string | undefined;
+
+  constructor(config: Config) {
+    this.endpoint = config.str("endpoint");
+    this.database = config.strOpt("database");
+  }
+
+  updateDatabase(database: string | undefined) {
+    if (database !== undefined) {
```
fauna-chase, line 204: what if I'm trying to unset the database?

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+      this.database = database;
+    }
+  }
+
+  makeSecret(endpoint?: Endpoint, scope?: string, role?: string) {
```
fauna-chase, line 209: we seem to be doing this secret generation in the endpoint and in the stack.  Would it be possible to have this logic just in the Endpoint class, and instead of calling this method we'd provide the endpoint method with the scope from the stack?

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+}): {
+  secret: string;
+  url: string;
+  graphqlHost: string;
+  graphqlPort: number;
+} => {
+  const flags = new Config("flag", opts.flags ?? {});
+
+  const rootConfig = new RootConfig(
+    new Config("config key", opts.rootConfig ?? {})
```
fauna-chase, line 286: nit: this obviously makes no difference at all... but I'd feel a bit better if our keys for these were something like `rootConfig` and `projectConfig`

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+  return lookupEndpoint0({
+    flags,
+    rootConfig,
+    localConfig,
+    scope,
+    role,
+  });
+};
+
+export const lookupEndpoint0 = (opts: {
```
fauna-chase, line 271: this is starting to feel like this is doing a lot more than just looking up and endpoint, which I suppose is the end result of what we are doing.  but we're also parsing and validating config.  I'm wondering if it would make sense to have something like a overarching Configuration class, that you create, and it parses and validates all your configuration.  Then on it there is a method to get the endpoint, and that, with everything already parsed, processed and validated no just does the things needed to return the endpoint.

this could take in all the ini config as well as the set of flags and build the correct config object.

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+  // An endpoint must be chosen as well. The endpoint may come from (in order)
+  // the `--endpoint` flag, the `endpoint` key in the local config, or the
+  // `default` key in the root config.
+  const endpointName =
+    flags.strOpt("endpoint") ?? stack?.endpoint ?? rootConfig.defaultEndpoint;
+
+  const secretFlag = flags.strOpt("secret");
+  if (endpointName === undefined && secretFlag === undefined) {
+    throw new Error(
+      "No endpoint or secret set. Set an endpoint in ~/.fauna-shell, .fauna-project, or pass --endpoint"
```
fauna-chase, line 333: would passing `--secret` also suffice here?

fauna-chase (commented):
functionality wise seems solid to me, most of my feedback is organizational.

A couple questions:
Do we require a default in the .fauna-project file?
what happens if you run CLI commands from a fauna project directory that doesn't have a default? would it just use the `~/.fauna-shell` default here? and if that didn't have one and you didn't specify one error?

I haven't looked into this too deeply yet, I'm assuming these files are just parsed every time a command is executed?

src/lib/config.ts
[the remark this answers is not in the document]
macmv: sure, I can rename to ProjectConfig.

My main hesitation with splitting this up is:
- its not all that large of a file. 400 lines seems perfectly reasonable to me
- typescript makes it difficult to re-export things. so if I defined `ProjectConfig` in `src/lib/config/project.ts`, it would be a bit ugly to write `import { ProjectConfig } from "src/lib/config"` elsewhere.

those are pretty minor though, I'll try reorganizing.

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+
+// Represents `.fauna-project`
+class LocalConfig {
+  defaultStack: string | undefined;
+  stacks: { [key: string]: Stack };
+
+  constructor(config: Config) {
+    this.defaultStack = config.strOpt("default");
+    this.stacks = Object.fromEntries(
+      config.entries((k) => k !== "default").map(([k, v]) => [k, new Stack(v)])
```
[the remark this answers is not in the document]
macmv: yes definitely

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+const ini = require("ini");
+
+class InvalidConfigError extends Error {}
+
+// Wraps an `ini` file with helpers to get typed values out
+class Config {
+  keyName: string;
+  private config: { [key: string]: unknown };
+
+  static parse(keyName: string, data: string) {
```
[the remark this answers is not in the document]
macmv: its not, I'll remove it. I'll see if I can get `eslint` to warn about unused functions like this.

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+      throw new InvalidConfigError(
+        `Expected number for ${this.keyName} ${key}, got ${typeof v}`
+      );
+    }
+  }
+
+  object(key: string): Config {
+    const v = this.config[key];
+    if (v !== null && typeof v === "object") {
+      return new Config(this.keyName, v as any);
```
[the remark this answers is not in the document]
macmv: no, because typescript is weird. `as` blindly changes the type of any value, regardless of what typescript thinks. But, the type `object` is not a subtype of the value needed for a config. Here's the error:
```ts
src/lib/config/index.ts:62:39 - error TS2345: Argument of type 'object' is not assignable to parameter of type '{ [key: string]: unknown; }'.
  Index signature for type 'string' is missing in type '{}'.

62       return new Config(this.keyName, v as object);
                                         ~~~~~~~~~~~
```

given that all objects implement the `[]` operator, I don't really understand this error, so for now imma just leave it at `as any`.

src/lib/config.ts
```diff
@@ -0,0 +1,417 @@
...
+    if (value === undefined) {
+      throw new InvalidConfigError(
+        `Missing value for required ${this.keyName} ${key}`
+      );
+    } else {
+      return value;
+    }
+  }
+
+  entries(pred: (key: string) => boolean): [string, Config][] {
```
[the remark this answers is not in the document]
macmv: yeah thats the intention. I renamed it to `allObjectsWhere`, which I think is clearer.

src/lib/config.ts
macmv: It would fail to parse. That file in JSON is like:
```json
{
  "default": "default",
  "default": {
    "secret": "blah",
  }
}
```

src/lib/config.ts
macmv: yeah I'll shorten it to that

src/lib/config.ts
macmv: the endpoint can update the url and graphql host/port, so I'm not sure that name would work

src/lib/config.ts
macmv: This function isn't used. The `database` field is public, so it should just be assigned to.

src/lib/config.ts
macmv: yeah that cleans things up nicely.

I guess this makes me wonder what the expected behavior of `fauna eval` is when you pass in a scope:
```
fauna my-db "2 + 3"
```
should `my-db` be appended to the project config database? or should it replace the project config database?

src/lib/config.ts
macmv: the keys there show up in errors. So for "config key", you'd see this error:
```
Expected string for config key url, got number
```
and for "flag" you'd see
```
Expected string for flag url, got number
```

so we could definitely update them to "root config key" and "project config key" or whatnot, but I don't think "rootConfig" and "projectConfig" would make that much sense.

src/lib/config.ts
macmv: Yeah, its a bit hard to pack all that into one error. The start says `No endpoint or secret` set, and I'd like to encourage people to use endpoints instead of passing `--secret`.

src/lib/config/project-config.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  constructor(config: Config) {
+    this.defaultStack = config.strOpt("default");
+    this.stacks = Object.fromEntries(
+      config.objectsIn("stack").map(([k, v]) => [k, new Stack(v)])
+    );
+
+    if (this.defaultStack === "default") {
+      throw new InvalidConfigError("Default stack cannot be named 'default'");
+    } else if (
```
fauna-chase, line 16: per offline discussion, I think this is valid for project config
```
default=default

[stack.default]
endpoint=cloud
database=default
```

fauna-chase (approved):

src/lib/config/project-config.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class Stack {
+  /**
+   * The endpoint name to use as a base.
+   */
+  endpoint: string;
+  /**
+   * The database path to use.
+   */
+  database?: string;
```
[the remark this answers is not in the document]
macmv: yeah

fauna-chase (approved):

fauna-chase (approved):

PR #248, opened 2023-09-14
Discover .fauna-project in parent directories

macmv:
Ticket(s): ENG-5535

Searches upwards from the current working directory for a `.fauna-project` file.

Things to note about this behavior:
- It will search upwards until it finds the root. So a `.fauna-project` in `/home` is valid, even if `.fauna-shell` is located in `/home/macmv`.
- It will immediately fail silently if fetching the CWD returns an error.
- It will ignore any directories or symlinks to directories named `.fauna-project`.
- It will consider the first file or symlink named `.fauna-project` "the project file." If this file is then unreadable, the shell will print an error and exit.

### DIFF
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -298,7 +298,7 @@ export const getProjectConfigPath = (): string | undefined => {
     return undefined;
   }
 
-  for (let i = 0; i < 2048; i++) {
+  while (true) {
     const projPath = path.join(current, ".fauna-project");
     let stat;
     stat = fs.statSync(projPath, {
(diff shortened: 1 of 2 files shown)


src/lib/config/index.ts
```diff
@@ -279,16 +279,51 @@ const readFileOpt = (fileName: string) => {
...
+  try {
+    current = process.cwd();
+  } catch (err) {
+    // If the cwd is not accessible, just give up. If this happens, one of our
+    // dependencies actually explodes elsewhere, but we might as well handle
+    // errors where possible.
     return undefined;
   }
+
+  for (let i = 0; i < 2048; i++) {
```
erickpintor, line 301: Does it need to be a bounded search? By definition, every filesystem I'm aware of will hit the `.base === ""` case eventually.

macmv: it doesn't need to be bounded. I was thinking it might be possible to create a loop with symlinks or something, but I can just make it unbounded.

erickpintor (approved):
LGTM

PR #249, opened 2023-09-14
Switch everything in lib/ to a module

macmv:
Ticket(s): ENG-5539

Because I added `config.ts`, we can't actually `require` it correctly. In dev, it needs to be `require("config.ts")`, whereas once its compiled, it needs to be `require("config.js")` (note the file extension changed). Using proper `import` statements lets us remove the file extension, which fixes the problem. Imports are only availible to modules though, so I went ahead and converted everything in `lib` to be a module. This also gives us the much clearer `export` keyword, instead of needing to use `module.exports`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,9 @@
           "@jest/globals"
         ]
       }
-    ]
+    ],
+    // we use typescript so dynamic imports are fine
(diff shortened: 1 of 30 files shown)


erickpintor (approved):
LGTM

PR #252, opened 2023-09-18
Store a ShellConfig in FaunaCommand

macmv:
Ticket(s): FE-###

Stores a `ShellConfig` in `FaunaCommand`, which will let us use the config in all the commands that need it.

### DIFF
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -99,30 +99,6 @@ export class Config {
   }
 }
 
-/**
- * Builds the options provided to the faunajs client.
- * Tries to load the ~/.fauna-shell file and read the default endpoint from there.
- *
- * Assumes that if the file exists, it would have been created by fauna-shell,
- * therefore it would have a defined endpoint.
- *
- * Flags like --host, --port, etc., provided by the CLI take precedence over what's
- * stored in ~/.fauna-shell.
- *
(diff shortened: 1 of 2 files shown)


src/lib/config/index.ts
```diff
@@ -245,10 +219,21 @@ export class ShellConfig {
...
   }
 
+  validate = () => {
+    if (this.endpoint === undefined) {
+      // No `~/.fauna-shell` was found, and no `--secret` was passed.
+      throw new Error(
+        "No endpoint or secret set. Set an endpoint in ~/.fauna-shell, .fauna-project, or pass --endpoint"
+      );
+    }
+  };
```
fauna-chase, line 229: we do almost all of the validation in the constructor for this class and then just do the endpoint in the validate method.  It seems that we are doing it this way because some commands don't require an endpoint.  Which ones are those?
Given that it seems we are saying that this class is technically 'valid' without an endpoint.  With that, I think I'd rather have the code paths that need it ensure there is an endpoint without this method on this class.

fauna-chase (approved):
other than my one minor comment, LGTM

src/lib/config/index.ts
macmv: yup, this should be renamed and refactored a bit to move more validation into here.

PR #254, opened 2023-09-25
Add searchable select input

macmv:
Ticket(s): ENG-5568

This will be used with `fauna stack add` and elsewhere, when selecting things.

### DIFF
--- a/src/lib/search-select.ts
+++ b/src/lib/search-select.ts
@@ -53,7 +53,11 @@ export const searchSelect = createPrompt(
     // TODO: Fuzzy filter?
     const items = choices.filter((item, _) => item.value.startsWith(typed));
 
+    // This is the item with a cursor next to it. The arrow keys will
+    // increment/decrement this value.
     const [activeState, setActive] = useState(0);
+    // This is the distance the amount that has been scrolled down. It is the
+    // index of the first visible item.
     const [scrollState, setScroll] = useState(0);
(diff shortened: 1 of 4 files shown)


src/lib/search-select.ts
```diff
@@ -0,0 +1,119 @@
...
+    const firstRender = useRef(true);
+    const prefix = usePrefix();
+    const [status, setStatus] = useState<"pending" | "done">("pending");
+    const [typed, setTyped] = useState("");
+
+    // TODO: Fuzzy filter?
+    const items = choices.filter((item, _) => item.value.startsWith(typed));
+
+    const [activeState, setActive] = useState(0);
+    const [scrollState, setScroll] = useState(0);
```
fauna-chase, line 57: on first read I'm struggling to understand what scroll is tracking, could you add a comment explaining that at a high level?

fauna-chase (approved):
overall LGTM, I think a more general concern would be the added code complexity vs the benefit for this.  If we only expect customers to have a few endpoints and a few stacks probably suffices just to have the arrow selection on the prompt then we don't need to worry about state management, paging, scrolling etc.

Given its complete and quite thoroughly tested I'm ok moving forward with it but always something to think about.

macmv:
this is true. i probably could have made due with something simpler, but I like me some gold plating :)

PR #255, opened 2023-09-26
Add `fauna stack add`

macmv:
Ticket(s): ENG-5533

Adds `fauna stack add`.

This takes a few arguments:
- `--name`: name of the new stack (optional)
- `--endpoint`: endpoint of the new stack (optional)
- `--database`: database of the new stack (optional)
- `--set-default`: sets the new stack as the default. (optional)
- `--non-interactive`: disables interaction, and makes `--name`, `--endpoint`, and `--database` required. (optional, defaults to false)

This commands asks you to select an endpoint, then select a database path, then choose a name for the stack. Finally, it asks you if you want to make the new stack the default. Each of these 4 questions will be skipped if a command line argument is given.

The first question, asking for an endpoint, will let you search and select an endpoint from `~/.fauna-shell`.

The second question, asking for a database path, will give you a sorted list of all database paths in that endpoint. This could be quite expensive, and you might want to set the database path to a database that doesn't exist yet. So this part needs a bit of work and thought for the expected behavior.

No questions will be asked if `--non-interactive` is given, and the `--name`, `--endpoint`, and `--database` flags will be required.

### DIFF
--- a/src/commands/stack/add.ts
+++ b/src/commands/stack/add.ts
@@ -1,0 +1,185 @@
+import { Command, Flags, ux } from "@oclif/core";
+import { input, confirm } from "@inquirer/prompts";
+import { Endpoint, ShellConfig } from "../../lib/config";
+import FaunaClient from "../../lib/fauna-client";
+import { searchSelect } from "../../lib/search-select";
+
+export default class AddStackCommand extends Command {
+  static flags = {
+    name: Flags.string({
+      description: "New stack name",
+    }),
+    endpoint: Flags.string({
+      description: "Endpoint to use in this stack",
+    }),
+    database: Flags.string({
+      description: "Database path to use in this stack",
+    }),
+    "non-interactive": Flags.boolean({
+      description: "Disables interaction",
+      dependsOn: ["name", "endpoint", "database"],
+    }),
+    "set-default": Flags.boolean({
+      description: "Sets this stack as the default",
+    }),
+  };
+
+  static description = "Adds a new stack to the `.fauna-project`.";
+
+  static examples = [
+    "$ fauna stack add",
+    "$ fauna stack add --name my-app --endpoint dev --database my-database",
+    "$ fauna stack add --name my-app --endpoint dev --database my-database --set-default",
+  ];
+
+  async run() {
+    const config = ShellConfig.read({});
+
+    await this.execute(config);
+  }
+
+  async execute(config: ShellConfig) {
+    const { flags } = await this.parse();
+
+    if (config.projectConfig === undefined) {
+      this.error(
+        "No `.fauna-project` found. Create one with `fauna project init`."
+      );
+    }
+
+    const endpointName =
+      flags.endpoint ??
+      (await searchSelect({
+        message: "Select an endpoint",
+        choices: Object.keys(config.rootConfig.endpoints).map((endpoint) => ({
+          value: endpoint,
+        })),
+      }));
+    if (!Object.keys(config.rootConfig.endpoints).includes(endpointName)) {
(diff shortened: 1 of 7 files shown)


src/commands/stack/add.ts
```diff
@@ -0,0 +1,185 @@
...
+    const name =
+      flags.name ??
+      (await input({
+        message: "Stack name",
+        validate: validateStackName,
+      }));
+    const res = validateStackName(name);
+    if (res !== true) {
+      this.error(res);
+    }
```
fauna-chase, line 89: since we are passing that method as a validate function to their input method, do we need to validate again?

src/commands/stack/add.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  promptDatabasePath = async (endpoint: Endpoint): Promise<string> => {
+    const { url, secret } = endpoint;
+    const client = new FaunaClient({ endpoint: url, secret });
+
+    const res = await client.query("0");
+    if (res.status != 200) {
+      this.error(`Error: ${res.body.error.code}`);
+    }
```
fauna-chase, line 118: are we doing this to validate the secret before we go grab the databases?

src/commands/stack/add.ts
```diff
@@ -0,0 +1,185 @@
...
+      }
+      if (path.length > depth_limit) {
+        overflowedDepth = true;
+        continue;
+      }
+
+      const databases = await client.query<string[]>(
+        `Database.all().take(${database_limit + 1}).toArray().map(.name)`,
+        {
+          secret: `${client.secret}:${path.join("/")}:admin`,
```
fauna-chase, line 159: do we have to put the admin here?

src/commands/stack/add.ts
```diff
@@ -0,0 +1,185 @@
...
+        }
+      );
+      if (databases.status !== 200) {
+        this.error(`Error: ${databases.body.error.code}`);
+      }
+
+      if (databases.body.data.length > database_limit) {
+        overflowedLimit = true;
```
fauna-chase, line 167: idt we currently use this anywhere, but as discussed offline, we want to rework the limit situation a bit, I went ahead and just created this jira:
https://faunadb.atlassian.net/browse/ENG-5574
to track that

src/lib/config/index.ts
```diff
@@ -252,6 +252,19 @@ export class ShellConfig {
 
     return this.endpoint!.makeScopedEndpoint(database, opts.role);
   };
+
+  /**
+   * Saves the project config, if present.
+   */
+  saveProjectConfig() {
+    this.projectConfig?.save(this.projectConfigFile()!);
+  }
```
fauna-chase, line 261: wondering if we should error if this is ever attempted without a project config?

fauna-chase (approved):

src/commands/stack/add.ts
macmv: we validate again in case they passed a command line argument

src/commands/stack/add.ts
macmv: yeah

src/commands/stack/add.ts
macmv: yeah, otherwise it'll parse the path as a role name

src/commands/stack/add.ts
macmv: perfect. this bool will be used later to determine if we should just let the user write in a path.

src/lib/config/index.ts
macmv: i don't really mind either way. the `?.` made it shorter to just do nothing if you don't have a project config, but I don't really have a preference.

PR #257, opened 2023-09-28
first pass at fauan project init

fauna-chase:
ENG-5534

### DIFF
--- a/src/commands/project/init.ts
+++ b/src/commands/project/init.ts
@@ -47,7 +47,9 @@ export class ProjectInitCommand extends Command {
       projectConfig: ProjectConfig.emptyConfig(),
     });
     const stackFactory = new StackFactory(this, shellConfig);
-    await stackFactory.addStack();
+    await stackFactory.addStack({
+      default: true,
+    });
   }
 
   private getProjectPath(projectDir?: string): string {
@@ -55,7 +57,7 @@ export class ProjectInitCommand extends Command {
       if (path.isAbsolute(projectDir)) {
         return projectDir;
       } else {
-        return `${path.join(process.cwd(), projectDir)}`;
+        return path.join(process.cwd(), projectDir);
       }
     } else {
       return process.cwd();
--- a/src/commands/project/init.ts
+++ b/src/commands/project/init.ts
@@ -1,0 +1,64 @@
+import { Args, Command } from "@oclif/core";
+import {
+  fileExists,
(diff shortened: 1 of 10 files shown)


src/lib/stack-factory.ts
```diff
@@ -0,0 +1,168 @@
...
+    this.cmd = cmd;
+    this.config = config;
+  }
+  /**
+   *
+   * @param cmd - used to send info/errors to the user.
+   * This method will be executed from the project init as well as stack add method
+   * @param config
+   */
+  async addStack(params?: AddStackParams) {
```
fauna-chase, line 29: this was all pulled fromt he AddStack command so that it can be used by that command as well as project init one.

src/commands/project/init.ts
```diff
@@ -0,0 +1,64 @@
...
+  ProjectConfig,
+  ShellConfig,
+} from "../../lib/config";
+import * as path from "path";
+import { StackFactory } from "../../lib/stack-factory";
+
+export class ProjectInitCommand extends Command {
+  static args = {
+    projectDir: Args.string({
+      name: "projectDir",
```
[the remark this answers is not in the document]
macmv: oh this is an arg not a flag, cool.

src/commands/project/init.ts
```diff
@@ -0,0 +1,64 @@
...
+    const stackFactory = new StackFactory(this, shellConfig);
+    await stackFactory.addStack();
+  }
+
+  private getProjectPath(projectDir?: string): string {
+    if (projectDir) {
+      if (path.isAbsolute(projectDir)) {
+        return projectDir;
+      } else {
+        return `${path.join(process.cwd(), projectDir)}`;
```
macmv, line 58:
```suggestion
        return path.join(process.cwd(), projectDir);
```

src/lib/config/index.ts
```diff
@@ -67,6 +69,11 @@ export class Config {
     }
   }
 
+  objectExists(key: string): boolean {
+    const v = this.config[key];
+    return v !== null && typeof v === "object";
```
macmv, line 74: I think this needs to be undefined, as `[]` will return undefined if there isn't a value present.
```suggestion
    return v !== undefined && typeof v === "object";
```

macmv (approved):
looks good, aside from a couple minor things.

macmv (approved):

PR #261, opened 2023-09-28
Add --endpointURL flag

macmv:
ENG-5530

Adds `--endpointURL`, and deprecates the old flags. I also chose to hide the old flags, as we really don't want new users to use them. So the old flags don't show up in `--help`, but they still exist and work as expected. Using the old flags will emit a warning though.

### DIFF
--- a/src/lib/config/root-config.ts
+++ b/src/lib/config/root-config.ts
@@ -88,21 +88,41 @@ export class Endpoint {
   }
 
   static getURLFromConfig = (config: Config): string | undefined => {
-    const url = config.strOpt("url");
-    const scheme = config.strOpt("scheme");
-    const domain = config.strOpt("domain");
-    const port = config.numberOpt("port");
+    return this.getURLInner({
+      url: config.strOpt("url"),
+      scheme: config.strOpt("scheme"),
+      domain: config.strOpt("domain"),
+      port: config.numberOpt("port"),
+    });
+  };
 
+  static getURLFromFlags = (flags: Config): string | undefined => {
+    return this.getURLInner({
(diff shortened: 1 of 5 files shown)


src/lib/config/root-config.ts
```diff
@@ -88,21 +88,41 @@ export class Endpoint {
...
+  static getURLFromFlags = (flags: Config): string | undefined => {
+    return this.getURLInner({
+      url: flags.strOpt("endpointURL"),
+      scheme: flags.strOpt("scheme"),
+      domain: flags.strOpt("domain"),
+      port: flags.numberOpt("port"),
+    });
+  };
 
+  static getURLInner = (opts: {
```
fauna-chase, line 108: nit: maybe just `getURL`? the Inner part makes me think there is some like nested url or something when I first read it.

fauna-chase (approved):
love that we can hide the params and they still work, nice!

src/lib/config/root-config.ts
```diff
@@ -88,21 +88,41 @@ export class Endpoint {
...
-    const port = config.numberOpt("port");
+    return this.getURLInner({
+      url: config.strOpt("url"),
+      scheme: config.strOpt("scheme"),
+      domain: config.strOpt("domain"),
+      port: config.numberOpt("port"),
+    });
+  };
+
+  static getURLFromFlags = (flags: Config): string | undefined => {
```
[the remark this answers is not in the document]
macmv: good point, i'll add a comment

src/lib/config/root-config.ts
macmv: makes sense, i'll rename it

PR #263, opened 2023-09-28
Cleanup add-endpoint

macmv:
Ticket(s): ENG-5532

Overhauls `fauna add-endpoint`. The old arguments/flags were too confusing, so I went ahead and replaced them:

- There is now a single optional argument, name, which is the new endpoint name.
- The flag `--url` has been added, which sets the endpoint url. I would be open to renaming this to `--endpointURL` to match `fauna eval` (optional, defaults to asking the user).
- The flag `--secret` has been added, which lets you set a secret (optional, defaults to asking the user).
- The flag `--set-default` has been added, which sets the new endpoint as default (optional, defaults to asking the user).
- The flag `--non-interactive` has been added, which disables any user interaction (for automated tooling). This makes `--url`, `--secret`, and the name argument required.

The previous `--alias` flag and `endpoint` argument have been removed, making this a breaking change. Given that this command was previously only usable with human interaction, I think its safe(ish) to break this.


src/commands/add-endpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+            const client = new FaunaClient({
+              secret: "secret",
+              endpoint: url,
+            });
+            try {
+              // This should fail with unauthorized. If it fails with another
+              // error, the URL is probably invalid.
+              await client.query(`0`);
+            } catch (e) {
+              return "Error: could not connect to URL";
```
fauna-chase, line 88: wondering if we should just validate the url here, in the scenario where the add a local host endpoint but don't have it running locally at the moment.

src/commands/add-endpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+      (await input({
+        message: "Database Secret",
+        validate: async (secret) => {
+          const client = new FaunaClient({ secret, endpoint: url });
+          const res = await client.query(`0`);
+          await client.close();
+          if (res.status !== 200) {
+            return `Error: ${res.body.error.code}`;
+          }
+          return true;
```
fauna-chase, line 110: same thing here with the validation, wondering if since we intend add endpoint to be mostly for local host dev, and we want to push people to cloud login for actual endpoints, may make sense to relax this validation a bit.

fauna-chase (approved):

src/commands/add-endpoint.ts
macmv: yup, agreed. i'll just validate the URL, then emit a warning if we can't connect to the DB

src/commands/add-endpoint.ts
macmv: agreed, i'll just emit a warning if we can't connect

PR #265, opened 2023-09-28
parse nested endpoint structure in root config

fauna-chase:
ENG-5544

### DIFF
--- a/src/lib/config/root-config.ts
+++ b/src/lib/config/root-config.ts
@@ -7,11 +7,29 @@ export class RootConfig {
 
   constructor(config: Config) {
     this.defaultEndpoint = config.strOpt("default");
-    this.endpoints = Object.fromEntries(
-      config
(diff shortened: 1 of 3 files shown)


src/lib/config/root-config.ts
```diff
@@ -26,6 +44,31 @@ export class RootConfig {
...
+   * will not be recognized.
+   */
+  private static configContainsNestedEndpointStructure(
+    config: Config
+  ): boolean {
+    if (config.objectExists("endpoint")) {
+      const endpointObj = config.object("endpoint");
+      return endpointObj.keys().some((key) => {
+        return endpointObj.objectExists(key);
+      });
```
macmv, line 67: style (you can avoid the second `return` if you don't make a block):
```suggestion
      return endpointObj.keys().some((key) => endpointObj.objectExists(key));
```

macmv (approved):

PR #268, opened 2023-10-03
Cleanup endpoint commands

macmv:
Ticket(s): ENG-5580

Adds a `fauna endpoint` topic, and updates commands to use the new shell config parsing.

New commands:
- `fauna endpoint add`. This is aliased to `fauna add-endpoint` for compatibility.
- `fauna endpoint list`. This is aliased to `fauna list-endpoints` for compatibility.
- `fauna endpoint remove`. This is aliased to `fauna delete-endpoint` for compatibility.


src/commands/endpoint/list.ts
```diff
@@ -0,0 +1,33 @@
+import { Command } from "@oclif/core";
+import { ShellConfig } from "../../lib/config";
+import chalk from "chalk";
+
+export default class ListEndpointCommand extends Command {
+  static flags = {};
+
+  static description = "Lists endpoints in ~/.fauna-shell.";
```
jdcryans, line 8: Looks like we typically don't use present tense in the descriptions, except for cloud-login. Please align.

src/commands/endpoint/remove.ts
```diff
@@ -0,0 +1,45 @@
+import { Args, Command } from "@oclif/core";
+import { ShellConfig } from "../../lib/config";
+
+export default class DeleteEndpointCommand extends Command {
+  static description = "Deletes an endpoint from ~/.fauna-shell.";
```
jdcryans, line 5: Same as above

src/commands/endpoint/list.ts
wdberkeley: We bounce back and forth between imperative and indicative mood. The schema commands and import, e.g., use imperative, but lots of others use indicative, like the delete-* and create-* commands. I prefer imperative: "Compute the frustum of the wangledinger", not "Computes the frustum of the wangledinger".

wdberkeley (approved):
LGTM. Would be nice to make the mood of the descriptions of all the commands be the same. Like I said in my comment, I prefer imperative, but indicate is OK.

src/commands/endpoint/list.ts
macmv: speaking of `delete-*`, do we want `fauna endpoint delete` or `fauna endpoint remove`? I added called it remove and noticed it was called delete beforehand. I'd be fine with either, with a very slight preference to `remove` (simply because all you're doing is removing a value from a config - you're not deleting any data).

macmv: we discussed in standup, and we want to go with `remove`. thats what is currently implemented, so leaving as-is.

PR #269, opened 2023-10-03
BT-4220: Update DIND Image

cynicaljoy:
Ticket(s): BT-4217

## Problem

Pipelines are failing when we use the latest version of the fauna docker image, but the issue doesn't appear to be with the image itself but rathe the version of Docker Engine that's used in `karlkfi/concourse-dcind`

## Solution

We forked `karlkfi/concourse-dcind` and created our own updated version of that image

## Result

Pipelines should be green 🤞 

## Testing

Details in Jira card -- TL;DR -- a bunch of Concourse ad-hoc runs that concluded this was a solution

### DIFF
--- a/concourse/tasks/integration-tests.yml
+++ b/concourse/tasks/integration-tests.yml
@@ -3,7 +3,10 @@ platform: linux
 image_resource:
   type: registry-image
   source:
-    repository: karlkfi/concourse-dcind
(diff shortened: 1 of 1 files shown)


macmv (approved):
the previous test failures were breaking in main, and now its fixed. looks good to me

PR #270, opened 2023-10-03
Re-implement cloud-login, using the new config

macmv:
Ticket(s): ENG-5578

Updates `cloud-login` to use the new ShellConfig class for updating the config.

Email/password, with MFA are supported. Login through github/netlify are not. This matches the behavior before this PR.

Note that this command is not tested, and we don't have a library to test user interaction, so I haven't added tests. I did enable MFA and test out all the things manually, and it seems to be working well.


src/commands/cloud-login.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (this.name == "global") {
+      return base;
+    } else {
+      return `${base}-${this.name}`;
+    }
+  }
+}
+
+export default class CloudLoginCommand extends Command {
+  static description = "Prompts the user to log into a Fauna account.";
```
jdcryans, line 34:
```suggestion
  static description = "Prompt the user to log into a Fauna account.";
```

jdcryans (approved):

src/commands/cloud-login.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (this.name === "global") {
+      return base;
+    } else {
+      return `${base}-${this.name}`;
+    }
+  }
+}
+
+export default class CloudLoginCommand extends Command {
+  static description = "Prompt the user to log into a Fauna account.";
```
wdberkeley, line 34: Minor: I think the description should say to the user what they will accomplish with the command, so this should be phrased as "Log in to a Fauna account".

wdberkeley (approved):

src/commands/cloud-login.ts
macmv: I like it, I'll update

PR #271, opened 2023-10-05
Fix scoped shell

macmv:
Ticket(s): FE-###

`ensureDbScopeClient` was completely busted. It didn't pass the `version` param through, and it didn't work with database paths. Both of these are fixed now.


README.md
```diff
@@ -1180,13 +1180,12 @@ Start an interactive shell.
...
-    <value>] [--secret <value>] [--endpoint <value>] [--file <value>] [--stdin]
-    [--output <value>] [--format json|json-tagged|shell] [--version 4|10]
-    [--typecheck]
+  $ fauna shell [SCOPE] [--endpointURL <value>] [--timeout <value>]
+    [--secret <value>] [--endpoint <value>] [--file <value>] [--stdin] [--output
+    <value>] [--format json|json-tagged|shell] [--version 4|10] [--typecheck]
 
 ARGUMENTS
-  DBNAME  database name
+  SCOPE  Database path
```
jdcryans, line 1188: Scope is a term we use internally, but I don't think it's going to represent something meaningful for users. FULLYQUALIFIEDPATH is a bit of a mouthful too. DB_PATH_AND_NAME?

README.md
```diff
@@ -1180,13 +1180,12 @@ Start an interactive shell.
 Start an interactive shell.
 
 USAGE
-  $ fauna shell [DBNAME] [--endpointURL <value>] [--timeout
-    <value>] [--secret <value>] [--endpoint <value>] [--file <value>] [--stdin]
-    [--output <value>] [--format json|json-tagged|shell] [--version 4|10]
-    [--typecheck]
+  $ fauna shell [SCOPE] [--endpointURL <value>] [--timeout <value>]
```
jdcryans, line 1183: Update the README?

README.md
macmv: I think `DB_PATH` or `DB_NAME` would work. its the same path shown in the dashboard on the current shell, so I think that has meaning. `DB_PATH_AND_NAME` is a bit long

README.md
macmv: yeah, I just updated the README to match the new `--help` output

jdcryans (approved):

PR #272, opened 2023-10-06
Remove catch all in performV10Query

macmv:
Ticket(s): FE-###

## Problem

## Solution

## Result

## Testing

### DIFF
--- a/src/commands/eval.js
+++ b/src/commands/eval.js
@@ -166,25 +166,21 @@ class EvalCommand extends FaunaCommand {
   }
 
   async performV10Query(client, fqlQuery, outputFile, flags) {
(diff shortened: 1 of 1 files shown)


macmv:
It looks like the first commit starts checking for duplicate object keys.

With just the first commit, performance remains unchanged (20ms for tagged, 40ms for interpolated).

With both commits, performance is greatly improved. This benchmark only tests values (no inline queries), so its the best case for this patch. I saw 20ms for tagged values, and 15ms with the interpolated object format.

The stddev listed by wrk is ± 1ms, and after restarting core a couple times, I saw about ± 2ms over all the runs.

PR #278, opened 2023-10-09
ENG-5593: Allow users to push non-logical changes

wdberkeley:
The CLI disallowed pushing if there were no logical changes unless --force was set, but it's perfectly reasonable to push formatting and comment changes only. Now this will be allowed.

I tested this manually. Here're samples of the workflow:

$ fauna schema push --dir=$HOME/fsl
Proposed diff:

foo.fsl add collection Bar
foo.fsl add collection Foo
Accept and push changes? [no]: yes

$ fauna schema push --dir=$HOME/fsl # Added a comment to Bar.
No logical changes to push
Push anyway? [no]: yes

$ fauna schema push --dir=$HOME/fsl
No logical changes to push
Push anyway? [no]: no
Push cancelled

### DIFF
--- a/src/commands/schema/push.js
+++ b/src/commands/schema/push.js
@@ -11,8 +11,8 @@ class PushSchemaCommand extends SchemaCommand {
     }),
   };
 
-  async confirm() {
-    const resp = await ux.prompt("Accept and push the changes?", {
+  async confirm(msg) {
+    const resp = await ux.prompt(msg, {
       default: "no",
     });
     if (["yes", "y"].includes(resp.toLowerCase())) {
@@ -22,7 +22,7 @@ class PushSchemaCommand extends SchemaCommand {
       return false;
     }
     console.log("Please type 'yes' or 'no'");
-    return this.confirm();
+    return this.confirm(msg);
   }
 
   async run() {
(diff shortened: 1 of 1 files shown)


src/commands/schema/push.js
```diff
@@ -52,13 +52,15 @@ class PushSchemaCommand extends SchemaCommand {
...
         }
-        if (!json.diff) {
-          this.log("No changes to push");
-          this.exit(0);
+        let msg = "Accept and push changes?"
+        if (json.diff) {
+          this.log(`Proposed diff:\n`);
+          this.log(json.diff);
+        } else {
+          this.log("No logical changes to push");
```
erickpintor, line 60: Idk if "logical" will resonate with users. Perhaps "No schema changes to push" or "No logical schema changes to push"?

wdberkeley: 🤔 I think this is a word in common developer usage that people will understand, either immediately or with a moment's thought. Maybe I'll change it to "No logical changes. Push new file contents anyway?". Does that sound better?

erickpintor: > No logical changes. Push new file contents anyway?

My minor suggestion on that is to drop the word "new" as we'll be pushing the whole file instead of just the "new bits". Other that that, it looks pretty good to me.

erickpintor (approved):
LGTM

PR #284, opened 2023-10-10
fix Connected to undefined message

fauna-chase:
ENG-5630
This will display the endpoint and database being connected to when ever a CLI command is executed.

```
p4 » local-shell schema diff
Connected to endpoint: cli_test-us database: cli_bugbash
No schema differences
```

```
p4 » local-shell shell
Connected to endpoint: cli_test-us database: cli_bugbash
Type Ctrl+D or .exit to exit the shell
>
```

### DIFF
--- a/src/lib/config/root-config.ts
+++ b/src/lib/config/root-config.ts
@@ -1,4 +1,4 @@
-import { Config, InvalidConfigError } from ".";
+import { Config, EndpointConfig, InvalidConfigError } from ".";
 import fs from "fs";
 const ini = require("ini");
 
@@ -23,13 +23,13 @@ export class RootConfig {
       this.endpoints = Object.fromEntries<Endpoint>(
         config
           .objectsIn("endpoint")
-          .map(([k, v]) => [k, Endpoint.fromConfig(v)])
+          .map(([k, v]) => [k, Endpoint.fromConfig(k, v)])
       );
     } else {
       this.endpoints = Object.fromEntries<Endpoint>(
         config
           .allObjectsWhere((k) => k !== "default")
-          .map(([k, v]) => [k, Endpoint.fromConfig(v)])
+          .map(([k, v]) => [k, Endpoint.fromConfig(k, v)])
       );
     }
 
@@ -101,14 +101,16 @@ export class RootConfig {
  * - An optional GraphQL port. This is the port to use for GraphQL requests. It defaults to `443`.
  */
 export class Endpoint {
+  name?: string;
   secret: string;
   url: string;
 
   graphqlHost: string;
   graphqlPort: number;
 
-  static fromConfig(config: Config) {
+  static fromConfig(name: string, config: Config) {
     return new Endpoint({
+      name: name,
       secret: config.str("secret"),
       url: Endpoint.getURLFromConfig(config),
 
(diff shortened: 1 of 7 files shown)


fauna-chase:
copying my jira comment here:

the previous impl was just printing the url it was connecting to.  I updated the current impl to do that, but do we think printing that url has any value?  for customers it will always be https://db.fauna.com or localhost feels like we could just strip it out entirely.

One thing that does seem valuable is printing the database that is being connected to.  Should we update it to do that?

macmv:
I like the idea of printing the current endpoint and database that its connected to. URL seems nice to have too, if we can fit all 3 of those in that one-liner.

src/lib/fauna-command.js
```diff
@@ -85,9 +84,7 @@ class FaunaCommand extends Command {
   mapConnectionError({ err, connectionOptions }) {
     if (err instanceof errors.Unauthorized) {
       return this.error(
-        `Could not Connect to ${stringifyEndpoint(
-          connectionOptions
-        )} Unauthorized Secret`
+        `Could not Connect to ${connectionOptions.url} Unauthorized Secret`
```
macmv, line 87:
```suggestion
        `Could not connect to ${connectionOptions.url}: unauthorized secret`
```

fauna-chase: hmm yeah I'm not sure that we want to do that, my bias would be to just leave it out

src/lib/config/root-config.ts
```diff
@@ -101,14 +101,16 @@ export class RootConfig {
  * - An optional GraphQL port. This is the port to use for GraphQL requests. It defaults to `443`.
  */
 export class Endpoint {
+  name?: string;
```
macmv, line 104:
```suggestion
  name: string;
```

src/lib/config/root-config.ts
```diff
@@ -101,14 +101,16 @@ export class RootConfig {
...
   secret: string;
   url: string;
 
   graphqlHost: string;
   graphqlPort: number;
 
-  static fromConfig(config: Config) {
+  static fromConfig(name: string, config: Config) {
     return new Endpoint({
+      name: name,
```
macmv, line 113: style:

```suggestion
      name,
```

src/lib/config/root-config.ts
```diff
@@ -118,41 +120,40 @@ export class Endpoint {
   }
 
   constructor(opts: {
+    name?: string;
```
macmv, line 123:
```suggestion
    name: string;
```

test/commands/endpoint.test.ts
```diff
@@ -114,6 +117,7 @@ describe("endpoint:add", () => {
           foobar: {
             url: "http://foo.baz",
             secret: "fn1234",
+            name: undefined,
```
macmv, line 120: can you update `name` to always be defined? it seems strange to only set it when reading the config, but not when writing it.

macmv (commented):
looks good to me, except for name being optional. IMO it should always be set.

fauna-chase:
> looks good to me, except for name being optional. IMO it should always be set.

if you provide a secret as a flag we create an endpoint without a name.

macmv:
oh yup nevermind

macmv (approved):

PR #285, opened 2023-10-10
Update schema prompts to inquerier

macmv:
Ticket(s): ENG-5602

This one is easier to read commit-by-commit.

It updates the prompts to use inquirer, and fixes database URLs that end in a slash, by constructing them with a `new URL`.


src/commands/schema/diff.ts
```diff
@@ -1,17 +1,20 @@
...
+import fetch from "node-fetch";
 
-class DiffSchemaCommand extends SchemaCommand {
+export default class DiffSchemaCommand extends SchemaCommand {
   static flags = {
     ...SchemaCommand.flags,
   };
 
+  static description = "Print the diff between local and remote schema.";
+  static examples = ["$ fauna schema diff --dir schemas/myschema"];
```
fauna-chase, line 10: should we show an example without the --dir as well?

src/commands/schema/push.ts
```diff
@@ -11,28 +12,18 @@ class PushSchemaCommand extends SchemaCommand {
...
-    }
-    if (["no", "n"].includes(resp.toLowerCase())) {
-      return false;
-    }
-    console.log("Please type 'yes' or 'no'");
-    return this.confirm(msg);
-  }
+  static description = "Push the current project's .fsl files to Fauna.";
+
+  static examples = ["$ fauna schema push --dir schemas/myschema"];
```
fauna-chase, line 17: same question with --dir here

src/commands/schema/push.ts
```diff
@@ -52,17 +43,23 @@ class PushSchemaCommand extends SchemaCommand {
...
           this.log(`Proposed diff:\n`);
           this.log(json.diff);
         } else {
           this.log("No logical changes.");
-          msg = "Push file contents anyway?"
+          message = "Push file contents anyway?";
         }
-        if (await this.confirm(msg)) {
+        this.log(`Proposed diff:\n`);
+        this.log(json.diff);
```
fauna-chase, line 55: are we logging the diff twice here?

fauna-chase (approved):

src/commands/schema/diff.ts
macmv: yeah we should

src/commands/schema/push.ts
macmv: oh yeah oops. let me go check, i think i messed up the rebase

fauna-chase:
o also, thank you for moving these all to type script :)

macmv:
oh but of course. it gives me nice auto-completions :)

PR #288, opened 2023-10-10
Update default-endpoint to `fauna endpoint select`

macmv:
Ticket(s): ENG-5620

Updates `fauna default-endpoint` to `fauna endpoint select`. I'm open to changing the name, this just lines up with `fauna stack select`.


src/commands/endpoint/select.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  async execute(config: ShellConfig) {
+    const { args } = await this.parse();
+    await this.parse();
+
+    const name =
+      args.name ??
+      (await searchSelect({
+        message: "Select a new default endpoint",
+        choices: Object.keys(config.rootConfig.endpoints).map((endpoint) => ({
```
fauna-chase, line 37: what happens if there are no endpoints here?

fauna-chase (approved):

src/commands/endpoint/select.ts
macmv: It won't work. I'll add a check for no endpoints, and error early and say something like "create an endpoin with `fauna cloud-login`"

PR #302, opened 2023-10-11
don't blow up for invalid config

fauna-chase:
ENG-5603

This will make it such that commands emit warnings when invalid endpoints are detected in the fauna shell config.  Users will still be able to use the shell if this is the case, all invalid endpoints will be ignored. As it currently stands, any attempt to update the fauna shell config file will fail if there are invalid endpoints as we would remove them in that scenario.

### DIFF
--- a/src/commands/stack/add.ts
+++ b/src/commands/stack/add.ts
@@ -31,7 +31,7 @@ export default class AddStackCommand extends Command {
   ];
 
   async run() {
-    const config = ShellConfig.read({});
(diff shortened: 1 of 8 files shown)


macmv (commented):
looks good, can you add a test in `config.test.ts` to ensure that `saveRootConfig` throws an error when given an invalid config?

macmv (approved):
oh i thought we could actually validate the creation of `invalidEndpoints`. i guess it isn't super important, lgtm

PR #308, opened 2023-10-13
Make --secret override project config

macmv:
Ticket(s): ENG-5650

I think this matches the feature matrix. I added some tests, and validated manually. We don't have mocks for the root/project config up, so I wasn't able to test all the combinations of things, but I'm mostly confident that its right :)

### DIFF
--- a/src/lib/secret.ts
+++ b/src/lib/secret.ts
@@ -1,12 +1,26 @@
 export class Secret {
   // A fauna key, like `fn1234`.
   key: string;
+  // Do we allow database scope?
+  allowDatabase: boolean;
   // A database scope, like `["foo", "bar"]`
   databaseScope: string[];
 
-  constructor(opts: { key: string; databaseScope?: string[] }) {
+  constructor(opts: {
+    key: string;
+    allowDatabase: boolean;
+    databaseScope?: string[];
+  }) {
     this.key = opts.key;
+    this.allowDatabase = opts.allowDatabase;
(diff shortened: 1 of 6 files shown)


src/lib/secret.ts
```diff
@@ -1,22 +1,36 @@
...
   }
 
+  static parseFlag(key: string) {
+    if (key.length === 0) {
+      throw new Error("Secret cannot be empty");
+    }
+    return new Secret({ key, allowDatabase: !key.includes(":") });
+  }
+
   static parse(key: string) {
```
fauna-chase, line 26: wondering if we should just have one parse method, could take as a parameter whether or not the key can be scoped.  Could use both of these methods under the hood.  looking at 2 methods parse vs parseFlag may not be clear which to use but if we had an explicit parameter, `allowScopedKey` or something, then it may make it a bit more clear.

fauna-chase (approved):

src/lib/secret.ts
```diff
@@ -1,22 +1,36 @@
...
     if (key.length === 0) {
       throw new Error("Secret cannot be empty");
     }
     if (key.includes(":")) {
       throw new Error("Secret cannot be scoped");
     }
```
fauna-chase, line 32: my understanding here is that we wouldn't allow scoped secrets in the `.fauna-shell` file.  I think that is fine, is backwards incompatible I believe though so could be a good thing for us to be aware of.

PR #318, opened 2023-10-18
add schema tests to the shell pipeline

fauna-chase:
I validated that the task works as expected with fly execute.  Once I verify it is working as expected in the pipeline I'll add it as a passed constraint to the publish job.  I'll also move the integ tests to the test job, that way we will always auto run the tests and publish can be executed with the latest version that passed the tests (or any selected version that passed the tests)

### DIFF
--- a/fsl/test-script.mjs
+++ b/fsl/test-script.mjs
@@ -1,0 +1,99 @@
+#!/usr/bin/env zx
+
+/**
+  * This is provided in the pipeline to pass in the built version of the
+  * fauna CLI to be tested.
+  */
+let faunaCmd;
+if (argv._.length === 1) {
+  faunaCmd = `${argv._[0]}`
+} else {
+  faunaCmd = "fauna"
+}
+
+const expectedCollNames = ["bye", "hi"]
+const expectedFuncNames = ["sayHello"]
+const secretFlag = process.env.FAUNA_SECRET
+if (secretFlag) {
+  $.verbose = false
+  console.log("A secret was found for FAUNA_SECRET, it will be used for all fauna CLI commands")
+}
+
+
+await ensureClean()
+
+/**
+  * Test Push
+  */
+await execFaunaCmd(["schema", "push", "--force"])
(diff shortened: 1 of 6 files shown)


fsl/test-script.mjs
```diff
@@ -0,0 +1,99 @@
...
+  */
+await execFaunaCmd(["schema", "push", "--force"])
+const collNames = await execPaginated("Collection.all().map(.name)")
+if (collNames.length != expectedCollNames.length && !collNames.every((elem, idx) => elem === expectedCollNames[idx])) {
+  throw new Error(`Schema collections do not match actual: ${collNames} expected: ${expectedCollNames}`)
+}
+const funcNames = await execPaginated("Function.all().map(.name)")
+if (funcNames.length != expectedFuncNames.length && !funcNames.every((elem, idx) => elem === expectedFuncNames[idx])) {
+  throw new Error(`Schema functions do not match actual: ${collNames} expected: ${expectedCollNames}`)
+}
```
macmv, line 36:
```suggestion
if (collNames.length != expectedCollNames.length || !collNames.every((elem, idx) => elem === expectedCollNames[idx])) {
  throw new Error(`Schema collections do not match actual: ${collNames} expected: ${expectedCollNames}`)
}
const funcNames = await execPaginated("Function.all().map(.name)")
if (funcNames.length != expectedFuncNames.length || !funcNames.every((elem, idx) => elem === expectedFuncNames[idx])) {
  throw new Error(`Schema functions do not match actual: ${collNames} expected: ${expectedCollNames}`)
}
```

macmv (approved):
looks good, except for the `||` comment.

might be nice to put this in the github PR ci too.

fauna-chase:
> looks good, except for the `||` comment.
> 
> might be nice to put this in the github PR ci too.

thats a good callout, I think I'll need to do a bit of tweaking there to use the docker image and create a fresh db and such so we don't run into any issues with them running concurrently.  Definitely seems useful to get them on the prs in that manner.

PR #324, opened 2023-11-03
stack -> environment

fauna-chase:
Also made our endpoint validation a bit less rigid.  This allows things like using fauna cloud-login when your project points to and endpoint that doesn't exist.  Any command that needs a valid endpoint still ensures it has one and will print an error message conveying that.

Added a custom help class so that we can display our beta topics below our non beta topics


README.md
```diff
@@ -516,15 +516,15 @@ Create a database.
...
+    [--secret <value>] [--endpoint <value>] [--environment <value>]
 
 ARGUMENTS
   DBNAME  database name
 
 FLAGS
   --endpoint=<value>  Connection endpoint, from ~/.fauna-shell
   --secret=<value>    Secret key. Overrides the `secret` in ~/.fauna-shell
-  --stack=<value>     Stack to use, from a Fauna project
+  --environment=<value>     environment to use, from a Fauna project
```
macmv, line 527: the README should be generated with the `readme.py` script, which will update this from the `--help` output.

fsl/.fauna-project
```diff
@@ -1,6 +1,6 @@
 schema_directory=schema
 default=dev
 
-[stack.dev]
+[environment.dev]
```
macmv, line 4: any plans to migrate the `stack` format to `environment`? I don't think its really worthwhile to add, I just realized that my fauna-shell file will be borked when I pull down this version.

src/commands/environment/select.ts
```diff
@@ -0,0 +1,43 @@
+import { Args, Command } from "@oclif/core";
+import { ShellConfig } from "../../lib/config";
+
+export default class SelectenvironmentCommand extends Command {
```
macmv, line 4:
```suggestion
export default class SelectEnvironmentCommand extends Command {
```

src/custom-help.ts
```diff
@@ -0,0 +1,44 @@
+import { Help, Interfaces } from "@oclif/core";
+
+/**
+ * This is here just so that we can place beta topics below non beta ones.  We extend the
+ * the Help class from oclif/core and just override their sortedTopics() method.
+ * https://github.com/oclif/core/blob/main/src/help/index.ts#L67
+ */
+export default class CustomHelp extends Help {
```
macmv, line 8: very cool, this change is nice to have.

macmv (commented):
looks good, just a couple minor changes

fsl/.fauna-project
fauna-chase: given we had only done the beta releases I was not planning to migrate them

macmv (approved):

PR #331, opened 2023-12-07
Write 0600 permissions when editing root config

macmv:
Ticket(s): FE-###

## Problem

~/.fauna-shell contains secrets, so it should not be readable by everyone by default.

## Solution

Write it with mode 0600.

## Result

It is not readable by others by default.

## Testing

I did `fauna cloud-login`, and the new file is only readable and writable by me:

```
$ ls -la ~/.fauna-shell*
-rw------- 1 macmv macmv  232 Dec  7 13:27 .fauna-shell
-rw-r--r-- 1 macmv macmv 1.5K Oct  9 10:49 .fauna-shell.bak
-rw-r--r-- 1 macmv macmv 1.5K Oct 12 15:40 .fauna-shell.old
```

### DIFF
--- a/src/lib/config/root-config.ts
+++ b/src/lib/config/root-config.ts
@@ -85,7 +85,9 @@ export class RootConfig {
     const config = this.toIni();
 
     const encoded = ini.encode(config);
-    fs.writeFileSync(path, encoded);
+    fs.writeFileSync(path, encoded, {
+      mode: 0o600,
+    });
   }
 
   toIni() {


fauna-chase (approved):

yesha-fauna:
Need to add warning if the fauna-shell or ProjectConfig Path does not have 600 permission.

src/lib/config/index.ts
```diff
@@ -329,9 +329,29 @@ export class ShellConfig {
         )}\n Resolve them by ensuring they have a secret defined or remove them if they are not needed.`,
         ...this.errors,
       ];
-    } else {
-      return [];
     }
+    if (!fileExistsWithPermission600(getRootConfigPath())) {
```
macmv, line 333: this check will log a warning if the file exists and is non-readable.

imo this check should be the other way around, and check for `fileExistsWithNon600Permissions`, so that only one check is needed, and the log won't show up for non-readable files and such.

macmv (commented):
looks good to me @yesha-fauna

PR #332, opened 2023-12-07
Bump to 1.2.1

macmv:
Ticket(s): FE-###

This release includes the switch over to `fetch`, and some improved error handling.

Seems like a patch to me, although I could be convinced it needs a minor version bump (just because the shell behavior changed slightly).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "fauna-shell",
   "description": "faunadb shell",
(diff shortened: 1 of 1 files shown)


fauna-chase (approved):
would like @cleve-fauna to validate that these changes resolved 1 and 2 from here for release if possible:
https://faunadb.atlassian.net/browse/FE-4795

PR #337, opened 2024-04-24
[DOCS-2723] Move project and environment commands to GA

jrodewig:
Ticket(s): [DOCS-2723](https://faunadb.atlassian.net/browse/DOCS-2724)

## Problem

Per @breinero, we're moving the project and environment functionality to GA.

## Solution

Updates the CLI's help topics to remove Beta flags and descriptions.

[DOCS-2723]: https://faunadb.atlassian.net/browse/DOCS-2723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


macmv (commented):
these docs are auto-generated with `yarn oclif readme`. what does it look like if you run that? i'm just curious, because that makes it easier to keep our code docs and readme in sync.

jrodewig:
> these docs are auto-generated with `yarn oclif readme`. what does it look like if you run that? i'm just curious, because that makes it easier to keep our code docs and readme in sync.

Thanks @macmv! Updated with 42cacca8bd41ff87f89a6565f670758e6010d71e.

macmv:
heh yeah i never said it was "good"

macmv (approved):
welp i'm not sure why so much has changed (oclif version maybe?), but it looks good to me.

macmv (changes requested):
oh actually, can you leave the `custom-help` line in the package.json? or remove the `custom-help.ts` file? I think @fauna-chase has more context on that, but it should either be kept as-is, or we can probably remove it.

jrodewig:
> oh actually, can you leave the `custom-help` line in the package.json? or remove the `custom-help.ts` file? I think @fauna-chase has more context on that, but it should either be kept as-is, or we can probably remove it.

Thanks for catching that. I was originally going to remove it, but decided to just get rid of the parts referencing projects and environments.

macmv (approved):
sounds good to me

PR #343, opened 2024-07-03
Improved error catching and logging

mwilde345:
Ticket(s): FE-5448

## Problem
* once core enforces v4 access (https://github.com/fauna/core/pull/11546), shell attempts to query v4 will result in 410's directly from fauna api
* Executing a v10 query within v4 client results in an unhandled error, a symptom of bad error catching

## Solution
* check for and display `err.message` and `err.description` from fauna. This will display the nice message from core regarding v4 access
* `eval` will properly catch errors from the async `wrapQueries` block

## Result
* Users see helpful error messages

## Testing
* added a test that mocks 410 response when setting up v4 client

### DIFF
--- a/src/lib/fauna-command.js
+++ b/src/lib/fauna-command.js
@@ -87,14 +87,14 @@ class FaunaCommand extends Command {
     }
   }
 
-  mapConnectionError({ err, connectionOptions, version }) {
+  mapConnectionError({ err, connectionOptions }) {
     if (err instanceof errors.Unauthorized) {
       this.error(
         `Could not Connect to ${connectionOptions.url} Unauthorized Secret`
       );
     } else {
-      const code = err?.message ? `${err.message}: ` : '';
-      const details = err?.description ?? '';
+      const code = err?.message ? `${err.message}: ` : "";
+      const details = err?.description ?? "";
       this.error(`${code}${details}`);
     }
   }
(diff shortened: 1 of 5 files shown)


echo-bravo-yahoo:
FYI, we'll probably respond with a 410 (that's what we return right now on account disablement).

src/lib/fauna-command.js
```diff
@@ -87,13 +87,22 @@ class FaunaCommand extends Command {
...
+    } else if (
+      err.description === "UnknownError" &&
+      err?.requestResult?.statusCode === 410 &&
+      version === "4"
+    ) {
+      this.error(
+        `This account is not allowed to query Fauna v4. Please use the v10 endpoint.`
+      );
+    } else {
+      this.error(err);
```
freels, line 104: This seems error prone, in that it assumes all unknown 410 errors are due to v4 being disabled. I think you should be able to get access to more error details via requestResult rather than relying on heuristics

mwilde345: Ashton and I tested locally and found no reason not to just forward the nice message from core: https://github.com/fauna/core/blob/33ae43bb9aeabdea0f596020b0514975020dcf9b/ext/api/src/main/scala/api/ExceptionResponseHelpers.scala#L68

Updated this block to just properly handle `description` and `message`

PR #345, opened 2024-07-12
Dependency upgrade: Minimum supported node version v18

echo-bravo-yahoo:
Ticket(s): [FE-5645](https://faunadb.atlassian.net/browse/FE-5645)

## Problem

Deprecation errors on every usage of the CLI

## Solution

This project's package.json engines field declared support all the way back to node 10, but we've taken dependencies that are incompatible with anything older than node 18! After updating the engines field, I did some digging and realized that native nodeJS fetch landed in v18, so we have an opportunity here to formalize our support boundary as node 18 and throw out the several fetch implementations we had lying around (one of which was responsible for the punycode deprecation warnings).

Also includes a .nvmrc to pin to v18 so we don't accidentally include libraries that require v19+ in the future.

## Result

```bash
➜  fauna-shell git:(punycode-deprecation-warnings) fauna shell
(node:11997) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Connected to endpoint: cloud
Type Ctrl+D or .exit to exit the shell
>
➜  fauna-shell git:(punycode-deprecation-warnings) node bin/run shell
Connected to endpoint: cloud
Type Ctrl+D or .exit to exit the shell
>
```

## Change Details

* Native fetch [requires](https://github.com/nodejs/node/issues/46221) addition of `duplex: "half"` attribute https://github.com/nodejs/node/issues/46221
* `nock` required upgrade to support native fetch interceptors https://github.com/nock/nock/issues/2397#issuecomment-1568646137
* `@oclif/test` required upgrade to get rid of the older `nock` dependency
  * substantial [migration changes](https://github.com/oclif/test/blob/main/MIGRATION.md) required
* New named export from `jest` for `expect` https://jest-archive-august-2023.netlify.app/docs/28.x/upgrading-to-jest28/#expect
* `@oclif/core` now exported as a module, moved out of `require` where necessary
* Moved [handle](https://github.com/oclif/core/blob/c925f6001c03b7a28c16ecbb8cdda0e48477cbed/guides/V3_MIGRATION.md#handle) `require` outside of the `catch` in `bin/run` as that fixed an issue where uncaught errors from OCLIF were having their entire stack trace printed.

## Testing
As we don't have 100% coverage, I went through and did some light manual testing for the following commands
- [x] cloud-login
- [x] add-endpoint
- [x] create-database
- [x] create-key
- [x] default-endpoint
- [x] delete-database
- [x] delete-endpoint
- [x] delete-key
- [x] eval
- [x] help
- [x] import
- [x] list-databases
- [x] list-endpoints
- [x] list-keys
- [x] run-queries
- [x] shell
- [x] upload-graphql-schema

[FE-5645]: https://faunadb.atlassian.net/browse/FE-5645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henryfauna (approved):
We should probably consider doing a major version bump when this goes out

henryfauna (commented):
You will also likely need to update the pipeline as it runs tests against node 16

echo-bravo-yahoo:
> You will also likely need to update the pipeline as it runs tests against node 16

Done!

mwilde345 (approved):

PR #346, opened 2024-07-22
v4 temporary header

mwilde345:
Ticket(s): FE-5644

## Problem

New users will eventually be unable to use the basic shell commands as they will not have v4 access

## Solution

Until all shell functions can be rewritten to use the v10 Fauna api, employ a temporary header that Core can use to continue to allow these v4 calls from the shell.

`shell` repl sessions and `eval` v4 commands will, however, not receive the temporary header and will work based on the user's access policy.

## Result

Unblock launch of v4 access restriction until we have bandwidth to rewrite shell functions to v10. 

Risk being that the header is publicly available and could be leveraged by any user regardless of their v4 access policy.

## Testing

Updated some tests to expect the header on commands other than `eval`. 

TODO: `shell` tests have been turned off for a while

### DIFF
--- a/src/lib/fauna-command.js
+++ b/src/lib/fauna-command.js
@@ -48,6 +48,16 @@ class FaunaCommand extends Command {
     super.error(message, { exit: 1 });
   }
 
+  _getHeaders() {
+    const headers = {
+      "X-Fauna-Source": "Fauna Shell",
+    };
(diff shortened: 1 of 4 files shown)


echo-bravo-yahoo (commented):
Tested locally w/ core and this branch, looks good to me!

echo-bravo-yahoo (approved):

PR #355, opened 2024-09-10
Add schema push --stage

macmv:
Ticket(s): ENG-6705

Adds `fauna schema push --stage`. This doesn't hit the right endpoint for the non-force version quite yet (as that endpoint doesn't exist), but everything else is in place to push a staged schema.

### DIFF
--- a/src/commands/schema/push.ts
+++ b/src/commands/schema/push.ts
@@ -9,6 +9,11 @@ export default class PushSchemaCommand extends SchemaCommand {
       description: "Push the change without a diff or schema version check",
       default: false,
     }),
+    stage: Flags.boolean({
+      description:
+        "Stages the schema change, instead of applying it immediately",
+      default: false,
+    }),
   };
 
   static description = "Push the current project's .fsl files to Fauna.";
@@ -16,6 +21,7 @@ export default class PushSchemaCommand extends SchemaCommand {
   static examples = [
     "$ fauna schema push",
     "$ fauna schema push --dir schemas/myschema",
+    "$ fauna schema push --stage",
   ];
 
   async run() {
@@ -24,8 +30,14 @@ export default class PushSchemaCommand extends SchemaCommand {
     try {
       const { url, secret } = await this.fetchsetup();
       if (this.flags?.force) {
-        // Just push.
-        const res = await fetch(new URL("/schema/1/update?force=true", url), {
+        const params = new URLSearchParams({
+          force: "true", // Just push.
+          staged: this.flags?.stage ? "true" : "false",
+        });
+
+        // This is how MDN says to do it for some reason.
+        const path = new URL(`/schema/1/update?${params}`, url);
+        const res = await fetch(path, {
           method: "POST",
           headers: { AUTHORIZATION: `Bearer ${secret}` },
           body: this.body(files),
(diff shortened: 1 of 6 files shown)


src/commands/schema/push.ts
```diff
@@ -34,21 +46,30 @@ export default class PushSchemaCommand extends SchemaCommand {
...
         }
       } else {
-        // Confirm diff, then push it.
-        const res = await fetch(new URL("/schema/1/validate?force=true", url), {
+        // Confirm diff, then push it. `force` is set on `validate` so we don't
+        // need to pass the last known schema version through.
+        const params = new URLSearchParams({
+          force: "true",
+        });
+        const path = new URL(`/schema/1/validate?${params}`, url);
```
freels, line 60: should we just drop the "force" param for validate?

freels (commented):
LGTM, would be good to get a DX team review

src/commands/schema/push.ts
macmv: you can pass a `version` instead, so we just require `force` or `version` to be set on all the schema endpoints. we could probably drop it, it just keeps the validation simple in the core endpoint.

freels: right, just thinking it might make using the API itself simpler. I forget/lost track if we're actually passing version to validate at any point

macmv: I need to double check, I think we might be using it in the dashboard to check if a schema version is stale. If we're not using it there, I'll go ahead and remove it from the core endpoint.

echo-bravo-yahoo:
Looks like CI is failing on linting:
```
/home/circleci/project/src/lib/config/index.ts
  196:14  warning  Constructor has a complexity of 23. Maximum allowed is 20  complexity

/home/circleci/project/src/lib/fauna-import-writer.js
  29:1  warning  Function 'getFaunaImportWriter' has too many parameters (6). Maximum allowed is 4  max-params
  ```

**Comment by macmv** [comment]
oh that's what its failing on! i couldn't figure that out for the life of me, all the tests looked fine

File: test/integ/schema.test.ts [hunk 2]
```diff
@@ -0,0 +1,52 @@
+import { expect } from "chai";
+import { cleanupDBs, evalOk, newDB, shellOk, stripMargin } from "./base";
+
+describe("fauna schema staged commands", () => {
+  // Cleanup after ourselves.
+  after(async function () {
+    await cleanupDBs();
+  });
+
+  it.only("fauna schema push --stage --force works", async () => {
```
echo-bravo-yahoo: remove `.only`

test/integ/schema.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        {
+          field: ".name",
+          mva: false,
+        },
+      ],
+      queryable: true,
+      status: "complete",
+    });
+
+    // But, the index should not be visible on the companion object.
```
echo-bravo-yahoo, line 41: +1 for comments with intent

echo-bravo-yahoo (approved):

macmv:
ok, tests are now failing because the container is a bit out of date. they should start passing next week.

mwilde345 (approved):

PR #356, opened 2024-09-10
Add fauna schema status

macmv:
Ticket(s): ENG-6711

Adds `fauna schema status`, which shows the current staged status for a database.

This accepts a `--dir` option, so that it'll integrate with the other project management stuff.

### DIFF
--- a/test/integ/schema.test.ts
+++ b/test/integ/schema.test.ts
@@ -51,4 +51,57 @@ describe.skip("fauna schema staged commands", () => {
       )
     ).to.deep.equal(null);
   });
+
+  const status = async (secret: string) => {
+    const output = await shellOk("fauna schema status --dir .", secret);
+
+    // Remove the "Connected to endpoint" line.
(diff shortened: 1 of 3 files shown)


test/integ/schema.test.ts
```diff
@@ -43,3 +43,56 @@ it("fauna schema push --stage --force works", async () => {
...
+       |        name: String
+       |        email: String
+       |    +
+       |    +   index byName {
+       |    +     terms [.name]
+       |    +   }
+       |      }
+       |
+       |
+       |`
```
freels, line 95: these are some interesting newlines...

freels (commented):
LGTM

test/integ/schema.test.ts
macmv: Yes indeed. I think two are from core, and one is from the shell.

mwilde345 (approved):

PR #364, opened 2024-09-23
New shell login setup

mwilde345:
Ticket(s): FE-5879

* Basic interface to call frontdoor auth code, token and session endpoints. 
* Only works locally as frontdoor only knows about the local client id.
* Fails if `new_session_tokens` is not turned on

### Out of Scope

In another PR i'll break some of this out to a new frontdoor client where I can add unit tests and other methods like refresh session

https://github.com/user-attachments/assets/4826c457-50f4-4044-a671-601beff799f8

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -41,12 +41,14 @@ export default class LoginCommand extends Command {
 
     const oAuth = new OAuthServer();
     await oAuth.start();
-    const dashboardOAuthURL = (await fetch(oAuth.getRequestUrl())).url;
-    const error = new URL(dashboardOAuthURL).searchParams.get("error");
-    if (error) {
-      throw new Error(`Error during login: ${error}`);
-    }
-    this.log(`To login, open your browser to:\n ${dashboardOAuthURL}`);
+    oAuth.server.on("ready", async () => {
+      const dashboardOAuthURL = (await fetch(oAuth.getRequestUrl())).url;
+      const error = new URL(dashboardOAuthURL).searchParams.get("error");
+      if (error) {
+        throw new Error(`Error during login: ${error}`);
+      }
+      this.log(`To login, open your browser to:\n ${dashboardOAuthURL}`);
+    });
     oAuth.server.on("auth_code_received", async () => {
       try {
         const tokenResponse = await oAuth.getToken();
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -37,7 +37,6 @@ export default class LoginCommand extends Command {
   }
 
   async execute() {
-    await this.parse();
 
     const oAuth = new OAuthServer();
     await oAuth.start();
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -9,7 +9,7 @@ type AccessToken = {
 };
 
 export default class LoginCommand extends Command {
-  static description = "Log in to a Fauna account.";
+  static description = "Login to your Fauna account.";
   static examples = ["$ fauna login"];
   static flags = {};
 
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -1,16 +1,5 @@
-import { input, confirm, password } from "@inquirer/prompts";
-import { ShellConfig } from "../lib/config";
-import { hostname } from "os";
 import { Command } from "@oclif/core";
-import { underline, blue } from "chalk";
 import OAuthServer from "../lib/auth/oauth-client";
-
-const DEFAULT_NAME = "cloud";
-const AUTH = process.env.FAUNA_AUTH ?? "https://auth.console.fauna.com";
-
-type Regions = {
-  [key: string]: Region;
-};
 
 type AccessToken = {
   access_token: string;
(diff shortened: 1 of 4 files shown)


src/lib/auth/oauth-client.ts
```diff
@@ -0,0 +1,170 @@
...
+    };
+    return `${accountURL}/api/v1/oauth/authorize?${new URLSearchParams(
+      params
+    )}`;
+  }
+
+  public getToken() {
+    const now = new Date();
+    // Short expiry for access token as it's only used to create a session
+    now.setUTCMinutes(now.getUTCMinutes() + 5);
```
mwilde345, line 54: @cleve-fauna should we enforce this in the backend where frontdoor looks at the clientID and overrides any ttl param with our own

mwilde345: probably since there's nothing stopping ppl from having a token that doesn't expire and continuing to create sessions

src/lib/auth/oauth-client.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  public async start() {
+    let port: number;
+    let isAvailable: boolean = false;
+
+    // Loop until an available port is found
+    do {
+      port = Math.floor(Math.random() * (63000 - 62500 + 1)) + 62500;
+      isAvailable = await this._isPortAvailable(port);
+    } while (!isAvailable);
```
echo-bravo-yahoo, line 155: thoughts on starting at a random point and walking linearly? this doesn't ever halt.

macmv: can you bind to port zero? that should let the OS pick a random port afaik

mwilde345: every time i tried on zero it just stuck with 0 🤷

mwilde345: that worked! https://github.com/fauna/fauna-shell/pull/364/commits/16c262cc3e0390e533649b7230268b882b9fe020

src/lib/auth/oauth-client.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  public async start() {
+    try {
+      this.port = await this._getFreePort();
+      this.server.listen(this.port);
+    } catch (e: any) {
+      console.error("Error starting loopback server:", e.message);
+    }
+  }
```
macmv, line 153: this will cause race conditions. the docs say a 'listening' event should be emitted once its ready, so i think something like this should work (that `event == 'listenting'` bit isn't right, but it should be close i think):
```suggestion
  private _listenOnFreePort(): Promise<number> {
    return new Promise((res, rej) => {
      this.server.listen(0, (event) => {
        if (event == 'listening') {
          this.port = (srv.address() as AddressInfo).port;
          res();
        }
      });
    });
  }

  public async start() {
    try {
      await this._listenOnFreePort();
    } catch (e: any) {
      console.error("Error starting loopback server:", e.message);
    }
  }
```

mwilde345: @macmv thanks I just committed something similar and it's working

macmv (commented):
nice, lgtm. i don't have much context on the oauth flow, so it'd be best to get someone else's approval too.

src/lib/auth/oauth-client.ts
```diff
@@ -0,0 +1,151 @@
...
+
+const accountURL = process.env.FAUNA_ACCOUNT_URL ?? "https://account.fauna.com";
+
+// Default to prod client id and secret
+const clientId = process.env.FAUNA_CLIENT_ID ?? "Aq4_G0mOtm_F1fK3PuzE0k-i9F0";
+// Native public clients are not confidential. The client secret is not used beyond
+//   client identification. https://datatracker.ietf.org/doc/html/rfc8252#section-8.5
+const clientSecret =
+  process.env.FAUNA_CLIENT_SECRET ??
+  "2W9eZYlyN5XwnpvaP3AwOfclrtAjTXncH6k-bdFq1ZV0hZMFPzRIfg";
```
cleve-fauna, line 14: I'd assume we're going to rip out these fallbacks eventually? (don't need to see here)

src/lib/auth/oauth-client.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    if (req.method === "GET") {
+      const parsedUrl = url.parse(req.url || "", true);
+      if (parsedUrl.pathname !== "/") {
+        errorMessage = "Invalid redirect uri";
+        this.closeServer();
+      }
+      const query = parsedUrl.query;
+      if (query.error) {
+        errorMessage = query.error.toString();
```
cleve-fauna, line 103: The authorization server in frontdoor also includes an error_description. Which would likely be helpful to the user.

src/commands/login.ts
```diff
@@ -0,0 +1,68 @@
...
+export default class LoginCommand extends Command {
+  static description = "Log in to a Fauna account.";
+  static examples = ["$ fauna login"];
+  static flags = {};
+
+  async run() {
+    await this.execute();
+  }
+
+  async getSession(access_token: string) {
```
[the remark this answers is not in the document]
mwilde345: correct

PR #365, opened 2024-09-24
Add colors

macmv:
Ticket(s): ENG-6785

Core is adding support for colors as query parameters from schema endpoints. This PR adds colors to all the schema commands.

This uses a library to check if colors are enabled, which will detect things like if we're in a TTY, and it'll detect common CI environments for us.

Additionally, new global `--color` and `--no-color` flags have been added. The library `supports-color` will parse `sys.argv`, and enable/disable colors accordingly. Because oclif makes it difficult to check if an optional flag is set, I'm going to leave this as-is, but it is currently quite brittle.

Testing has been manual, oclif doesn't provide a way to unit test with colors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "object-sizeof": "^1.6.1",
     "prettier": "^2.3.0",
     "rate-limiter-flexible": "^2.3.6",
-    "stream-json": "^1.7.3"
+    "stream-json": "^1.7.3",
+    "supports-color": "^8"
   },
   "devDependencies": {
     "@babel/eslint-parser": "^7.17.0",
@@ -39,6 +40,7 @@
     "@types/mocha": "^10.0.1",
     "@types/node": "^20.6.0",
     "@types/sinon": "^17.0.3",
+    "@types/supports-color": "^8.1.3",
     "@typescript-eslint/parser": "6.7.3",
     "c8": "^8.0.1",
     "chai": "^4.2.0",
--- a/src/lib/color.ts
+++ b/src/lib/color.ts
@@ -1,0 +1,28 @@
+import { stdout } from "supports-color";
+
+var colorEnabled: boolean | undefined = undefined;
+export const disableColor = () => {
+  colorEnabled = false;
+};
+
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -29,7 +29,8 @@
     "object-sizeof": "^1.6.1",
     "prettier": "^2.3.0",
     "rate-limiter-flexible": "^2.3.6",
-    "stream-json": "^1.7.3"
+    "stream-json": "^1.7.3",
+    "supports-color": "^8"
```
macmv, line 33: also, `supports-color` 9 requires ESM, which we aren't using. so, like oclif, we're going to use this older version.

oclif doesn't expose the result of "supports-color", so we need to depend on this separately.

src/lib/color.ts
```diff
@@ -0,0 +1,10 @@
+import { stdout } from "supports-color";
+
+// The value for the `color` parameter to the `/schema/1` endpoints.
+export const colorParam = (): string => {
+  return hasColor() ? "ansi" : "";
+};
+
+export const hasColor = (): boolean => {
+  return stdout !== false && stdout.hasBasic;
```
macmv, line 9: `supports-color` defines 4 levels of color:
- no support
- basic support (0-16 basic ansi colors)
- 256 color support
- full color support

Core only emits the basic 0-16 ansi colors, so we just check for that.

src/lib/fauna-command.js
```diff
@@ -299,6 +299,15 @@ FaunaCommand.flags = {
   environment: Flags.string({
     description: "Environment to use, from a Fauna project",
   }),
+
+  // FIXME(neil): Currently, the library `supports-color` parses out this flag,
+  // and enables or disables collor accordingly. This is somewhat horrible, but
+  // plumbing this flag down to all the commands would be more of a pain.
+  // So I'm going to leave it as-is, but this is quite brittle.
```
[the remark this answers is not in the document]
macmv: oh i missed the `init` block, yeah i'll rework this.

freels (approved):

PR #367, opened 2024-09-26
Add fauna schema diff --active

macmv:
Ticket(s): ENG-6796

Currently, `fauna schema diff` diffs against the active schema, which isn't what I'd expect it to do. This changes the default to diff against the staged schema, adds a flag to toggle the behavior, and shows which schemas are being compared in the command output.

### DIFF
--- a/src/commands/schema/diff.ts
+++ b/src/commands/schema/diff.ts
@@ -1,9 +1,14 @@
 import SchemaCommand from "../../lib/schema-command";
-import { colorParam, hasColor } from "../../lib/color";
+import { bold, colorParam, hasColor, reset } from "../../lib/color";
+import { Flags } from "@oclif/core";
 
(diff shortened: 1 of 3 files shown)


src/commands/schema/diff.ts
```diff
@@ -1,9 +1,14 @@
...
 export default class DiffSchemaCommand extends SchemaCommand {
   static flags = {
     ...SchemaCommand.flags,
+    active: Flags.boolean({
+      description: "Compare the local schema to the active schema.",
+      default: false,
+    }),
   };
 
   static description = "Print the diff between local and remote schema.";
```
echo-bravo-yahoo, line 14: thoughts on clarifying like:
`Compare the local schema to the active schema instead of the staged schema.`

and

`Print the diff between the local schema and the staged remote schema.`

echo-bravo-yahoo (approved):

PR #370, opened 2024-09-30
login poc

mwilde345:
Copy of https://github.com/fauna/fauna-shell/pull/364 that was closed, against new `v3` branch changes

Ticket(s): FE-5879

* Basic interface to call frontdoor auth code, token and session endpoints. 
* Only works locally as frontdoor only knows about the local client id.
* Fails if `new_session_tokens` is not turned on

### Out of Scope

In another PR i'll break some of this out to a new frontdoor client where I can add unit tests and other methods like refresh session

https://github.com/user-attachments/assets/4826c457-50f4-4044-a671-601beff799f8


src/lib/auth/oauth-client.ts
```diff
@@ -0,0 +1,156 @@
...
+export const ACCOUNT_URL =
+  process.env.FAUNA_ACCOUNT_URL ?? "https://account.fauna.com/api/v1";
+
+// Default to prod client id and secret
+const clientId = process.env.FAUNA_CLIENT_ID ?? "Aq4_G0mOtm_F1fK3PuzE0k-i9F0";
+// Native public clients are not confidential. The client secret is not used beyond
+//   client identification. https://datatracker.ietf.org/doc/html/rfc8252#section-8.5
+const clientSecret =
+  process.env.FAUNA_CLIENT_SECRET ??
+  "2W9eZYlyN5XwnpvaP3AwOfclrtAjTXncH6k-bdFq1ZV0hZMFPzRIfg";
```
echo-bravo-yahoo, line 15: is this really not confidential? what are the defaults used for / do they work for accounts besides yours?

if they are confidential, we should rewrite history to remove references to these (and probably delete & recreate the relevant secret, too).

src/commands/login.ts
```diff
@@ -0,0 +1,84 @@
+import { Command } from "@oclif/core";
+import OAuthServer, { ACCOUNT_URL } from "../lib/auth/oauth-client";
+import open from "open";
```
echo-bravo-yahoo, line 3: want to pair on rewriting this to yargs and adding a few tests? the relevant bits are in the v3 branch.

src/commands/login.ts
```diff
@@ -0,0 +1,84 @@
...
+      headers: myHeaders,
+    };
+    const response = await fetch(`${ACCOUNT_URL}/databases`, requestOptions);
+    if (response.status >= 400) {
+      throw new Error(`Error listing databases: ${response.statusText}`);
+    }
+    const databases = await response.json();
+    console.log(databases);
+    return databases;
+  }
```
echo-bravo-yahoo, line 52: is this reusable by the list databases command? if so, should we extract it into a helper?

mwilde345: not reusable, just added quick for the demo. Going to break it out when working on https://faunadb.atlassian.net/browse/FE-5876

src/lib/auth/oauth-client.ts
mwilde345: yeah we leave the secret in here to adhere to the api input shape. But it doesn't really serve a purpose
https://datatracker.ietf.org/doc/html/rfc8252#section-8.5

We do what else we can. Enforcing redirect URI's match (localhost)

Similar to github CLI: https://github.com/cli/cli/blob/4f0264e37b2b42253ea866f07b010e5b840aa416/internal/authflow/flow.go#L21

Nothing is stopping someone from distributing their own app with our client ID to try and impersonate Fauna CLI. 

Using PKCE, state, and checking authorization code for a matching client ID help protect from MITM, but that's true for public and confidential clients.

mwilde345:
adding tests in another pr

PR #371, opened 2024-10-01
add schema push

echo-bravo-yahoo:


src/lib/schema.mjs
```diff
@@ -0,0 +1,76 @@
...
+    throw new Error(`The project fsl directory: ${dir} is not writeable.`);
+  }
+}
+
+// Reads the files using their relative-to-`basedir` paths and returns their
+// contents paired with the relative path.
+// Fails if the total size of the files is too large.
+// relpaths: string[]
+function read(dir, relpaths) {
+  const FILESIZE_LIMIT_BYTES = 32 * 1024 * 1024;
```
macmv, line 19: files are documents in the database, so you can limit this to 8mb.

src/lib/schema.mjs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Gathers all FSL files in the directory rooted at `basedir` and returns a
+// list of relative paths.
+// Fails if there are too many files.
+// returns string[]
+async function gatherRelativeFSLFilePaths(dir) {
+  const logger = (await container.resolve("logger"))
+
+  const FILE_LIMIT = 256;
```
macmv, line 43: I'd raise this to like 32k or something. There's no reason why someone wouldn't want a whole bunch of schema files, as splitting things into multiple files should be effectively free.

src/yargs-commands/schema/push.mjs
```diff
@@ -0,0 +1,122 @@
...
+        params.set("staged", "true")
+
+      const path = new URL(`/schema/1/update?${params}`, argv.url);
+      const res = await fetch(path.toString(), {
+        method: "POST",
+        headers: { AUTHORIZATION: `Bearer ${argv.secret}` },
+        body: fsl,
+        // https://github.com/nodejs/node/issues/46221
+        // https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1483
+        // @ts-expect-error-next-line
```
macmv, line 26: nit: is this compiled with typescript? or is it just javascript?

src/lib/schema.mjs
```diff
@@ -0,0 +1,76 @@
...
+      const isDir = fs.statSync(fp).isDirectory();
+      if (n.endsWith(".fsl") && !isDir) {
+        curr.push(relp);
+      }
+      if (isDir) {
+        subdirs.push(relp);
+      }
+    }
+    for (const reldir of subdirs) {
+      curr.concat(go(reldir, curr));
```
macmv, line 60: might want to check the length while recursing, so that recursive symlinks don't blow up the program.

src/yargs-commands/schema/push.mjs
```diff
@@ -0,0 +1,122 @@
...
+        method: "POST",
+        headers: { AUTHORIZATION: `Bearer ${argv.secret}` },
+        body: fsl,
+        // @ts-expect-error-next-line
+        duplex: "half",
+      });
+
+      const json = await res.json();
+      if (json.error) {
+        logger.stderr(json.error?.message ?? json.error);
```
macmv, line 51: if this fails, the command should exit. does `logger.stderr` cause the command to exit immediately?

macmv (commented):
just a few nitpicks. also, i'm curious: does the new CLI use typescript, or is it all vanilla JS?

mwilde345 (approved):

echo-bravo-yahoo:
> just a few nitpicks. also, i'm curious: does the new CLI use typescript, or is it all vanilla JS?

currently, the new CLI is vanilla ESM. no build, no bundler.

i expect we'll add esbuild before release for distribution reasons and lean on JSDoc => intellisense for code assistance in the repo.

src/yargs-commands/schema/push.mjs
echo-bravo-yahoo: raw js - the typescript annotations don't do anything, but this was copied from `src/commands/schema/push.ts`, so a few of the non-breaking TS flavored bits stuck around. i'll take these out now while i'm thinking about them.

src/lib/schema.mjs
echo-bravo-yahoo: i'm planning on a wholesale rewrite of the fs traversal - going to leave as-is until then.

src/yargs-commands/schema/push.mjs
echo-bravo-yahoo: it does not - the schema commands right now are very WIP.

PR #375, opened 2024-10-04
Persist Account keys on login

mwilde345:
Ticket(s): FE-5875

## Problem

We aren't persisting account keys

## Solution

Build basic handler to manage creds files

* Credentials class has methods to get, put, and delete entries in the `access_keys` or `secret_keys` file. 
* `profile` flag can be used to index the stored account keys. `default` is otherwise used.
* added some jsdocs

## Result

Basic persistence of creds

Made a simple list databases handler `fauna db list` that uses the stored account key. Will build that command and tests out later.

## Testing

* Login successfully writes and overwrites account key creds

### DIFF
--- a/src/lib/auth/oauth-client.mjs
+++ b/src/lib/auth/oauth-client.mjs
@@ -12,7 +12,6 @@ const clientSecret =
 const REDIRECT_URI = `http://127.0.0.1`;
 
 class OAuthClient {
-
   constructor() {
     this.server = http.createServer(this._handleRequest.bind(this));
     this.code_verifier = Buffer.from(randomBytes(20)).toString("base64url");
@@ -118,7 +117,7 @@ class OAuthClient {
     try {
       if (!this.server.listening) {
         this.server.on("listening", () => {
-          this.port = (this.server.address()).port;
+          this.port = this.server.address().port;
           this.server.emit("ready");
         });
         this.server.listen(0);
--- a/src/lib/auth/oauth-client.mjs
+++ b/src/lib/auth/oauth-client.mjs
@@ -12,17 +12,6 @@ const clientSecret =
 const REDIRECT_URI = `http://127.0.0.1`;
 
 class OAuthClient {
-  server; //: http.Server;
-
-  port; //: number;
-
-  code_verifier; //: string;
-
-  code_challenge; //: string;
-
-  auth_code; //: string;
-
-  state; //: string;
 
   constructor() {
     this.server = http.createServer(this._handleRequest.bind(this));
@@ -127,11 +116,11 @@ class OAuthClient {
 
   async start() {
     try {
-      this.server.on("listening", () => {
-        this.port = this.server.address().port;
-        this.server.emit("ready");
-      });
       if (!this.server.listening) {
+        this.server.on("listening", () => {
+          this.port = (this.server.address()).port;
+          this.server.emit("ready");
+        });
         this.server.listen(0);
       }
     } catch (e) {
--- a/src/lib/fauna-account-client.mjs
+++ b/src/lib/fauna-account-client.mjs
@@ -96,7 +96,7 @@ export class FaunaAccountClient {
     headers.append("Authorization", `Bearer ${accessToken}`);
 
     const requestOptions = {
-      method: "GET",
+      method: "POST",
       headers,
     };
     try {
--- a/src/lib/fauna-account-client.mjs
+++ b/src/lib/fauna-account-client.mjs
@@ -2,13 +2,35 @@
 
 import { container } from "../cli.mjs";
 
+/**
+ * Class representing a client for interacting with the Fauna account API.
+ */
 export class FaunaAccountClient {
+  /**
+   * Creates an instance of FaunaAccountClient.
+   */
   constructor() {
+    /**
+     * The base URL for the Fauna account API.
+     * @type {string}
+     */
     this.url =
       process.env.FAUNA_ACCOUNT_URL ?? "https://account.fauna.com/api/v1";
+
+    /**
+     * The fetch function for making HTTP requests.
+     * @type {Function}
+     */
     this.fetch = container.resolve("fetch");
   }
 
+  /**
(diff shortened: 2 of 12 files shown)


src/lib/auth/oauth-client.mjs
```diff
@@ -12,18 +12,6 @@ const clientSecret =
 const REDIRECT_URI = `http://127.0.0.1`;
 
 class OAuthClient {
```
mwilde345, line 14: will jsdoc this in another pr, this one is getting big

echo-bravo-yahoo: in the JSDoc PR can you add the annotation `//@ts-check` to the top of the file? it'll turn on in-IDE static analysis and should help you generate good types.

src/lib/auth/oauth-client.mjs
```diff
@@ -127,11 +115,11 @@ class OAuthClient {
...
-      this.server.on("listening", () => {
-        this.port = this.server.address().port;
-        this.server.emit("ready");
-      });
       if (!this.server.listening) {
+        this.server.on("listening", () => {
+          this.port = this.server.address().port;
+          this.server.emit("ready");
+        });
         this.server.listen(0);
```
echo-bravo-yahoo, line 123: is `this.server.listening` set when you emit `ready`?

src/lib/fauna-account-client.mjs
```diff
@@ -21,6 +43,18 @@ export class FaunaAccountClient {
...
 
+  /**
+   * Retrieves an access token from the Fauna account API.
+   *
+   * @param {Object} opts - The options for the token request.
+   * @param {string} opts.clientId - The client ID for the OAuth application.
+   * @param {string} opts.clientSecret - The client secret for the OAuth application.
+   * @param {string} opts.authCode - The authorization code received from the OAuth authorization.
+   * @param {string} opts.redirectURI - The redirect URI for the OAuth application.
+   * @param {string} opts.codeVerifier - The code verifier for the OAuth PKCE flow.
```
echo-bravo-yahoo, line 54: are any of these optional? if so, please use the JSDoc optional annotation, e.g.:
```
@param {string} [opts.codeVerifier] - The code verifier for the OAuth PKCE flow.
```

if not, disregard

src/lib/fauna-account-client.mjs
```diff
@@ -66,22 +106,26 @@ export class FaunaAccountClient {
...
+   *
+   * @param {string} accountKey - The account key to list databases for.
+   * @returns {Promise<Object[]>} - The list of databases.
+   * @throws {Error} - Throws an error if there is an issue during the request.
+   */
+  async listDatabases(accountKey) {
     const headers = new Headers();
-    headers.append("Authorization", `Bearer ${account_key}`);
+    headers.append("Authorization", `Bearer ${accountKey}`);
+
```
echo-bravo-yahoo, line 128: thoughts on abstracting out something with an API like `makeFaunaRequest` on the next draft of this PR? that'll let you centralize error handling and make these methods really short & similar.

src/lib/file-util.mjs
```diff
@@ -27,3 +29,136 @@ export function dirIsWriteable(path) {
...
+    if (!fileExists(this.filepath)) {
+      fs.writeFileSync(this.filepath, "{}");
+    }
+  }
+
+  /**
+   * Retrieves the value associated with the given key from the credentials file.
+   *
+   * @param {string} [key] - The key to retrieve the value for.
+   * @returns {any} - The value associated with the key, or the entire parsed content if no key is provided.
```
echo-bravo-yahoo, line 75: let's add a `@typedef` for the credentials object? it can start with something very generic, e.g., `/** @typedef {Object.<string, any>} credentialsObject */` and we can refine it, but this way we're at least passing around a named type that asserts it's not undefined.

src/lib/file-util.mjs
```diff
@@ -27,3 +29,136 @@ export function dirIsWriteable(path) {
...
+      // Open file for reading and writing without truncating
+      const fileContent = fs
+        .readFileSync(this.filepath, { flag: "r+" })
+        .toString();
+      if (!isJSON(fileContent)) {
+        this.logger.stderr(
+          "Credentials file contains invalid formatting: ",
+          this.filepath
+        );
+        this.exit(1);
```
echo-bravo-yahoo, line 88: prefer to throw an error over logging & exiting (and note that the error constructor only takes one string), e.g.:
```
throw new Error(`Credentials file contains invalid formatting: ${this.filepath}`)
```

src/lib/file-util.mjs
```diff
@@ -27,3 +29,136 @@ export function dirIsWriteable(path) {
...
+      }
+      const parsed = JSON.parse(fileContent);
+      return key ? parsed[key] : parsed;
+    } catch (err) {
+      this.logger.stderr(
+        "Error while parsing credentials file: ",
+        this.filepath,
+        err
+      );
+      this.exit(1);
```
echo-bravo-yahoo, line 98: same - throw an error!

src/lib/file-util.mjs
```diff
@@ -27,3 +29,136 @@ export function dirIsWriteable(path) {
...
+      const existingContent = overwrite ? {} : this.get();
+      const newContent = {
+        ...existingContent,
+        [profile]: creds,
+      };
+      fs.writeFileSync(this.filepath, JSON.stringify(newContent, null, 2));
+    } catch (err) {
+      this.logger.stderr("Error while saving credentials: ", err);
+      this.exit(1);
+    }
```
echo-bravo-yahoo, line 121: throw

src/lib/auth/oauth-client.mjs
mwilde345: this check was a workaround for that builtYargs.argv weirdness, so not strictly necessary but yes by the time i send "ready" it's already `true`

echo-bravo-yahoo (approved):

PR #379, opened 2024-10-07
add tscheck annotations, add JSDoc types to makeFaunaRequest

echo-bravo-yahoo:
Ticket(s): FE-###

## Problem

## Solution

## Result

## Testing

### DIFF
--- a/src/lib/command-helpers.mjs
+++ b/src/lib/command-helpers.mjs
@@ -1,3 +1,5 @@
+//@ts-check
+
 function buildHeaders() {
   const headers = {
     "X-Fauna-Source": "Fauna Shell",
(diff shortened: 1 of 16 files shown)


src/lib/command-helpers.mjs
```diff
@@ -14,10 +16,11 @@ export async function getSimpleClient(argv) {
     const faunadb = (await import("faunadb")).default;
     const { Client, query: q } = faunadb;
     const { hostname, port, protocol } = new URL(argv.url);
+    const scheme =  protocol?.replace(/:$/, ""),
     client = new Client({
       domain: hostname,
-      port,
-      scheme: protocol?.replace(/:$/, ""),
+      port: Number(port),
+      scheme: /** @type {('http'|'https')} */ (scheme),
```
mwilde345, line 23: is this the best way to type it? Can we type the `Client`

mwilde345 (approved):

src/lib/command-helpers.mjs
echo-bravo-yahoo: The v4 driver client has the type `'http' | 'https' | undefined`, which seems pretty accurate to me. I don't think we want to loosen that type.

mwilde345: oh was mostly referring to the syntax/inline comment way of doing things

PR #380, opened 2024-10-09
Improve diff args

macmv:
Ticket(s): ENG-6821

The following arguments to `fauna schema diff`:
- <no args>: returns the diff from staged to local schema.
- `staged`: returns the diff from active to staged schema.
- `active`: returns the diff from active to local schema.

### DIFF
--- a/src/commands/schema/diff.ts
+++ b/src/commands/schema/diff.ts
@@ -1,14 +1,23 @@
 import SchemaCommand from "../../lib/schema-command";
 import { bold, colorParam, hasColor, reset } from "../../lib/color";
-import { Flags } from "@oclif/core";
+import { Args, Flags } from "@oclif/core";
+
+type Target = "local" | "staged" | "active";
 
 export default class DiffSchemaCommand extends SchemaCommand {
   static flags = {
     ...SchemaCommand.flags,
-    active: Flags.boolean({
-      description:
-        "Compare the local schema to the active schema instead of the staged schema.",
+    text: Flags.boolean({
+      description: "Display the text diff instead of the semantic diff.",
       default: false,
+    }),
+  };
+
+  static args = {
+    ...SchemaCommand.args,
+    target: Args.string({
+      description: "The target schema to compare against (active | staged).",
+      required: false,
     }),
   };
 
@@ -20,6 +29,8 @@ export default class DiffSchemaCommand extends SchemaCommand {
   ];
 
   async run() {
+    const [source, target] = this.parseTarget();
+
     const fps = this.gatherRelativeFSLFilePaths();
     const files = this.read(fps);
     const { url, secret } = await this.fetchsetup();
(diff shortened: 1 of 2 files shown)


freels (commented):
LGTM, we can tweak it based on feedback in the next few days.

src/commands/schema/diff.ts
```diff
@@ -1,17 +1,27 @@
...
+      description: "Display the text diff instead of the semantic diff.",
       default: false,
     }),
   };
 
+  static args = {
+    ...SchemaCommand.args,
+    target: Args.string({
+      description:
+        "The target schema to compare against, or a source and target in the form <source>..<target>.",
```
jrodewig, line 20:
```suggestion
        "The target schema to compare against.",
```

jrodewig (commented):
Just a minor comment

src/commands/schema/diff.ts
macmv: ah yeah i missed that, thanks.

mwilde345 (approved):

src/commands/schema/diff.ts
```diff
@@ -20,76 +29,107 @@ export default class DiffSchemaCommand extends SchemaCommand {
...
+    }
+
+    if (target === "active") {
+      return ["active", "local"];
+    } else if (target === "staged") {
+      return ["active", "staged"];
+    } else {
+      // NB: `this.error` quits the program, so return `any` to make typescript happy.
+      this.error("Invalid target. Expected: active or staged");
+      return [] as any;
```
echo-bravo-yahoo, line 132: if you expand the try/catch block in `run()` to include the `parseTarget()` call, you can throw from here instead of exiting with `this.error` and satisfy the type contract in a more consistent way.

src/commands/schema/diff.ts
```diff
@@ -1,17 +1,26 @@
...
+type Target = "local" | "staged" | "active";
 
 export default class DiffSchemaCommand extends SchemaCommand {
   static flags = {
     ...SchemaCommand.flags,
-    active: Flags.boolean({
-      description:
-        "Compare the local schema to the active schema instead of the staged schema.",
+    text: Flags.boolean({
+      description: "Display the text diff instead of the semantic diff.",
```
echo-bravo-yahoo, line 11: out of curiosity, what do these 2 formats look like?

src/commands/schema/diff.ts
```diff
@@ -1,17 +1,26 @@
...
+    ...SchemaCommand.args,
+    target: Args.string({
+      description: "The target schema to compare against (active | staged).",
+      required: false,
+    }),
+  };
+
   static description =
     "Print the diff between local schema and staged remote schema.";
   static examples = [
```
echo-bravo-yahoo, line 26: this is probably worth adding ~2 examples, one using a non-default `target` flag, and one using a non-default `diffKind`.

echo-bravo-yahoo (approved):

PR #383, opened 2024-10-09
don't delete fsl files if user rejects confirmation prompt

echo-bravo-yahoo:


src/commands/schema/diff.ts
```diff
@@ -97,16 +97,14 @@ export default class DiffSchemaCommand extends SchemaCommand {
           this.log(
             `Differences from the ${bold()}remote${reset()} schema to the ${bold()}local${reset()} schema:`
           );
+        } else if (source === "active") {
+          this.log(
+            `Differences from the ${bold()}remote, active${reset()} schema to the ${bold()}local${reset()} schema:`
+          );
```
echo-bravo-yahoo, line 103: eslint suggested simplifying this conditional - feel like i agree.

PR #387, opened 2024-10-11
Switch push to default to a staged push

macmv:
Ticket(s): ENG-6837 ENG-6833 ENG-6835

Staged push is the way!

### DIFF
--- a/src/commands/schema/push.ts
+++ b/src/commands/schema/push.ts
@@ -10,9 +10,9 @@ export default class PushSchemaCommand extends SchemaCommand {
       description: "Push the change without a diff or schema version check",
       default: false,
     }),
-    staged: Flags.boolean({
+    "skip-staged": Flags.boolean({
       description:
-        "Stages the schema change, instead of applying it immediately",
+        "Skip staging the schema and make the schema active immediately. This will cause indexes to be temporarily unavailable.",
       default: false,
     }),
   };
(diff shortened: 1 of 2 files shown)


src/commands/schema/push.ts
```diff
@@ -10,9 +10,9 @@ export default class PushSchemaCommand extends SchemaCommand {
       description: "Push the change without a diff or schema version check",
       default: false,
     }),
-    staged: Flags.boolean({
+    unstaged: Flags.boolean({
       description:
-        "Stages the schema change, instead of applying it immediately",
+        "Skips the staged schema. This will cause indexes to be temporarily unavailable.",
```
echo-bravo-yahoo, line 15: how about something with more emphasis on the verb and outcome? e.g.,: `Skip staging the schema and make the schema active immediately. This will cause indexes to be temporarily unavailable.`

src/commands/schema/push.ts
```diff
@@ -10,9 +10,9 @@ export default class PushSchemaCommand extends SchemaCommand {
       description: "Push the change without a diff or schema version check",
       default: false,
     }),
-    staged: Flags.boolean({
+    unstaged: Flags.boolean({
```
echo-bravo-yahoo, line 13: i think it might be more elegant to do one of:
- `--skip-staging` / `--skip-staged` (boolean, optional, default false)
- `--target stage` / `--target staged` / `--target active` (string, optional, default stage)

macmv: I like `--skip-staged,` I'll update.

src/commands/schema/push.ts
macmv: Good idea, I'll update.

echo-bravo-yahoo (approved):

PR #392, opened 2024-10-13
Add `fauna schema pull --active`

macmv:
Ticket(s): ENG-6832

Adds `fauna schema pull --active`. This toggles the `?staged` query parameter. Both `--staged` and `--active` are disallowed when there is no staged schema.

### DIFF
--- a/src/commands/schema/pull.ts
+++ b/src/commands/schema/pull.ts
@@ -27,27 +27,11 @@ export default class PullSchemaCommand extends SchemaCommand {
     const { url, secret } = await this.fetchsetup();
 
     try {
-      // Gather remote schema files to download.
-      const filesres = await fetch(new URL("/schema/1/files", url), {
+      // Check if there's a staged schema, and require `--staged` if there is one.
+      const statusres = await fetch(new URL(`/schema/1/staged/status`, url), {
         method: "GET",
         headers: { AUTHORIZATION: `Bearer ${secret}` },
       });
-      const filesjson = await filesres.json();
-      if (filesjson.error) {
-        this.error(filesjson.error.message);
-      }
-
-      // Check if there's a staged schema, and require `--staged` if there is one.
-      const params = new URLSearchParams({
-        version: filesjson.version,
-      });
-      const statusres = await fetch(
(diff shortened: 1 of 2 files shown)


src/commands/schema/pull.ts
```diff
@@ -13,7 +13,13 @@ export default class PullSchemaCommand extends SchemaCommand {
       default: false,
     }),
     staged: Flags.boolean({
-      description: "Pulls staged schema instead of the active schema",
+      description:
+        "Pulls the staged schema. Not applicable when there is no staged schema.",
```
freels, line 17: This description is misleading IMO. I think the intuition we want developers to have is that there is effectively always an active and staged schema pointer, but they can point to the same thing.

This flag should be the default.

src/commands/schema/pull.ts
```diff
@@ -13,7 +13,13 @@ export default class PullSchemaCommand extends SchemaCommand {
...
+      default: false,
+    }),
+    active: Flags.boolean({
+      description:
+        "Pulls the active schema. Not applicable when there is no staged schema.",
```
freels, line 22: Similar comment as above.

src/commands/schema/pull.ts
```diff
@@ -27,36 +33,50 @@ export default class PullSchemaCommand extends SchemaCommand {
...
+        if (this.flags?.staged && this.flags?.active) {
+          this.error(
+            "`--active` and `--staged` cannot be passed at the same time."
+          );
+        } else if (!this.flags?.staged && !this.flags?.active) {
+          this.error(
+            "There is a staged schema. Pick the staged or active schema with `--staged` or `--active`."
+          );
+        } else if (this.flags?.staged) {
+          isStagedPull = true;
```
freels, line 64: This API is too brittle: No flag combo works unconditionally, they all error out sometimes based on db state.

I think instead you can make `--staged` the default, always pulling the latest schema. The stdout output can communicate clearly what is happening.

Alternatively, (contradicting myself a bit) maybe it's just `--staged` which errors out: `--active` always pulls the active and the default pulls whatever is the latest. That feels a bit inconsistent, though.

mwilde345 (approved):

PR #393, opened 2024-10-14
persist secrets

mwilde345:
Ticket(s): FE-###

## Problem

## Solution

## Result

## Testing


src/cli.mjs
```diff
@@ -102,6 +115,11 @@ function buildYargs(argvInput) {
         default: [],
         choices: ["fetch", "error", "argv"],
       },
+      // Whether authNZ middleware should run. Better way of doing this?
+      authRequired: {
+        hidden: true,
+        default: false,
+      },
```
echo-bravo-yahoo, line 122: what's the intent here / why would you change it?

src/commands/key.mjs
```diff
@@ -1,16 +1,34 @@
...
+  // TODO: when using fauna to create a key at the specified database path, we should
+  //  getDBKey(parent path).
+  const dbSecret = await getDBKey({
+    accountKey,
+    path: database,
+    role,
+    url,
+  });
+  console.log("got account key", accountKey);
+  console.log("got db secret", dbSecret);
```
echo-bravo-yahoo, line 31: use verbosity-based logging (logger.debug etc)

src/lib/auth/authNZ.mjs
```diff
@@ -0,0 +1,192 @@
...
+    if (database) {
+      await setDBKey({ accountKey, path: database, role, url });
+    }
+  } catch (e) {
+    // Should prompt login when
+    // 1. Account key is not found in creds file
+    // 2. Account key is found but it and refresh token are invalid/expired
+    if (e instanceof InvalidCredsError) {
+      promptLogin(profile);
+    } else {
```
echo-bravo-yahoo, line 31: should skip prompt (and fail) if `argv.noInput` is set

echo-bravo-yahoo: ah, i see that prompt just bails. this is ok with `noInput` set or unset.

src/lib/auth/authNZ.mjs
```diff
@@ -0,0 +1,192 @@
...
+    throw e;
+  }
+}
+
+export async function checkDBKeyRemote(dbKey, url) {
+  const client = new FaunaClient({
+    secret: dbKey,
+    endpoint: url,
+  });
+  const result = await client.query("0");
```
echo-bravo-yahoo, line 181: is there any advantage to pre-checking instead of trying to do the work? is it that we can do it still in the middleware before reaching the handler?

echo-bravo-yahoo (approved):

PR #394, opened 2024-10-14
Simplify flags to push and diff

macmv:
Ticket(s): ENG-6842

Renames `push --skip-staged` to `push --active`, and switches `diff` to take flags instead of an argument.


echo-bravo-yahoo:
is the end state of this change that `schema push` and `schema diff` both use a flag `--active|--staged` that defaults to `--staged`?

mwilde345 (approved):

PR #397, opened 2024-10-15
Rename --force and --non-interactive to --no-input

macmv:
Ticket(s): FE-###

Renames `--force` and `--non-interact` to `--no-input`.


freels (approved):
LGTM. We can tweak help descs later if necessary.

PR #404, opened 2024-10-22
Add Github action for testing

echo-bravo-yahoo:
Ticket(s): FE-6016

## Problem
We currently do build/test in CircleCI, out of reach of any open source contributors.

## Solution
Test in a github action.

## Result
Contributors have a better experience w/viewing test results.


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
+
+name: Test
+
+on:
+  push:
+    branches: [ "main", "v3" ]
+  pull_request:
+    branches: [ "main", "v3" ]
```
echo-bravo-yahoo, line 10: the "v3" branch inclusion is temporary while we're working on the next version of the CLI

ecooper: Do we want to run this and circle on main while we're working out of v3?

ecooper (approved):
LGTM

.github/workflows/node.js.yml
echo-bravo-yahoo: good point - let me turn it off for main

PR #409, opened 2024-10-25
automated SEA builds

echo-bravo-yahoo:
Ticket(s): FE-6028

## Problem

We shouldn't be building SEA binaries by hand.

## Solution

Let's build as many SEA binaries as possible in github actions.

## Result

We can build these binaries in github actions:
- windows x64
- mac x64
- mac arm64 (M1, unsure if compatible with M2)
- linux x64

This misses some OS x arch setups that may be important (e.g., linux arm64), but those will be harder to do in github actions.

### DIFF
--- a/sea/build.cjs
+++ b/sea/build.cjs
@@ -18,7 +18,9 @@ if (platform === "linux") {
   console.log(`Building for Windows (${arch})...`);
   buildSEAForWindows();
 } else {
-  throw new Error(`No build configured for platform ${platform} and arch ${arch}!`);
+  throw new Error(
+    `No build configured for platform ${platform} and arch ${arch}!`,
+  );
 }
 
 function buildSEAForLinux() {
--- a/sea/build.cjs
+++ b/sea/build.cjs
@@ -1,0 +1,59 @@
+/* eslint-disable no-console */
+
+const os = require("node:os");
+const { execSync: execSyncActual } = require("node:child_process");
+const fs = require("node:fs");
+const execSync = (command, options) =>
+  execSyncActual(command, { ...options, encoding: "utf-8" });
+
+const platform = os.platform();
+const arch = os.arch();
+if (platform === "linux") {
+  console.log(`Building for Linux (${arch})...`);
+  buildSEAForLinux();
+} else if (platform === "darwin") {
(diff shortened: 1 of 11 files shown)


sea/build.cjs
```diff
@@ -0,0 +1,57 @@
...
+if (platform === "linux") {
+  console.log("Building for Linux...");
+  buildSEAForLinux();
+} else if (platform === "darwin") {
+  console.log("Building for Mac...");
+  buildSEAForMac();
+} else if (platform === "win32") {
+  console.log("Building for Windows...");
+  buildSEAForWindows();
+}
```
echo-bravo-yahoo, line 19: should throw here...

sea/build.cjs
```diff
@@ -0,0 +1,57 @@
...
+
+const os = require("node:os");
+const { execSync } = require("node:child_process");
+const fs = require("node:fs");
+const run = (command, options) =>
+  execSync(command, { ...options, encoding: "utf-8" });
+
+const platform = os.platform();
+if (platform === "linux") {
+  console.log("Building for Linux...");
```
echo-bravo-yahoo, line 11: should include the arch in these log lines

sea/build.cjs
```diff
@@ -0,0 +1,57 @@
...
+  // more details on signing:
+  // https://learn.microsoft.com/en-us/dotnet/framework/wcf/feature-details/how-to-create-temporary-certificates-for-use-during-development#installing-a-certificate-in-the-trusted-root-certification-authorities-store
+  // const signtool = "C:\\\"Program Files (x86)\"\\\"Microsoft SDKs\"\\ClickOnce\\SignTool\\signtool.exe";
+  // run(`${signtool} remove /s /c /u .\\dist\\fauna.exe`);
+  run(
+    "npx postject .\\dist\\fauna.exe NODE_SEA_BLOB .\\dist\\sea.blob ^ \
+    --sentinel-fuse NODE_SEA_FUSE_fce680ab2cc467b6e072b8b5df1996b2",
+  );
+  // run(`${signtool} sign /fd SHA256 .\\dist\\fauna.exe`);
+  run("chmod +x ./dist/fauna");
```
echo-bravo-yahoo, line 56: what. this gotta go.

PR #415, opened 2024-11-18
login redirect

mwilde345:
Ticket(s): FE-6132

## Problem
The shell automatically follows the oauth response's redirect. Doing a fetch against the resulting dashboard url returns a 401. 

* `FaunaAccountClient` class methods were all static and couldn't be called from the helpers

## Solution

* Rather than setting `shouldThrow` to `false` on the `makeAccountRequest` call, just get the url in 2 steps.
  * Don't follow redirects in the fetch call
  * Get the redirect target from the response headers
* Opted to keep `FaunaAccountClient` around as a class even though it just serves as a collection of helpers to call the account api, no class variables are necessary. However I moved `makeAccountRequest` to be an injectable module. In the future I want to mock the 'edge' only, i.e. calls to `makeAccountRequest` instead of mocking the class methods in `FaunaAccountClient`.

## Result
No unexpected error during login flow
Screenshot 2024-11-19 at 12 28 34 PM

## Testing

`npm run test:local`


echo-bravo-yahoo (approved):

.gitignore
```diff
@@ -14,3 +14,4 @@ experiments
 .log
 coverage
 test-results.xml
+.history
```
echo-bravo-yahoo, line 17: 👍

echo-bravo-yahoo (approved):

PR #416, opened 2024-11-19
Shell command

echo-bravo-yahoo:
Ticket(s): FE-6124

Implements the `shell` functionality of the old CLI.

Additionally:
- adds stdout/stderr to the DI container to allow tests to assert on the whole output.


test/helpers.mjs
```diff
@@ -29,3 +32,63 @@ export function buildUrl(path, paramObject) {
...
+      console.log(`Byte ${i}: ${charCode1}, ${charCode2}`);
+  }
+}
+
+/**
+ * Class representing a no-frills writable stream that, instead of writing data
+ * to a destination, holds it in memory. Can be queried for this data later;
+ * use it for testing interfaces that use streams.
+ */
+export class InMemoryWritableStream extends Writable {
```
echo-bravo-yahoo, line 70: Test writable stream. For readable streams, I've just been using `Readable.from()` and then using the native methods on it.

test/login.mjs
```diff
@@ -90,16 +92,14 @@ describe("login", function () {
     expect(oauthClient.start.called).to.be.true;
     // We open auth url in the browser and prompt user
     expect(container.resolve("open").calledWith("dashboard-url"));
-    expect(
-      logger.stdout.calledWith(
```
echo-bravo-yahoo, line 94: This syntax actually can't fail - gotta move the calledWith outside the expect.

src/config/setup-test-container.mjs
```diff
@@ -49,24 +55,7 @@ export function setupTestContainer() {
...
-      error: logger.error,
-      fatal: logger.fatal,
-
-      // use these for communicating with customers.
-      // mocked because they _are_ part of our public
-      // interface and should be tested.
-      stdout: stub(),
-      stderr: stub(),
-    }),
+    logger: awilix.asFunction((cradle) => spy(buildLogger(cradle))).singleton(),
```
echo-bravo-yahoo, line 58: I really like this pattern - spy on building a singleton injectable.

src/commands/shell.mjs
```diff
@@ -0,0 +1,109 @@
...
+    preview: argv.version !== "10",
+    // TODO: integrate with fql-analyzer for completions
+    completer: argv.version === "10" ? () => [] : undefined,
+    output: container.resolve("stdoutStream"),
+    input: container.resolve("stdinStream"),
+    eval: await buildCustomEval(argv),
+    terminal: true,
+  };
+
+  const shell = repl.start(replArgs);
```
echo-bravo-yahoo, line 32: like the prior CLI, this uses node's REPL, which gives us lots of thing for free (syntax highlighting in some cases, standard keyboard shortcuts, prompts, etc).

mwilde345 (approved):

PR #417, opened 2024-11-19
persist secrets

mwilde345:
Ticket(s): FE-6132

## Problem
* No tests to validate the behavior of the authNZ middleware

## Solution

* add them. 
  * Includes a combination of stubbing out some class methods (e.g. `AccountKey.get`) and also stubbing out the 'edge' (`fetch` and `fs`)

## Result

* We know what functionality the authNZ middleware provides

## Testing
Screenshot 2024-11-20 at 1 17 16 PM


package.json
```diff
@@ -66,7 +66,7 @@
     "pretest": "npm run format && npm run lint",
     "lint": "eslint . --fix",
     "test": "mocha --recursive ./test --require ./test/mocha-root-hooks.mjs --reporter spec --reporter mocha-junit-reporter",
-    "test:local": "mocha --recursive ./test --require ./test/mocha-root-hooks.mjs",
+    "test:local": "mocha --recursive ./test/authNZ.mjs --require ./test/mocha-root-hooks.mjs",
```
echo-bravo-yahoo, line 69: remove!

src/cli.mjs
```diff
@@ -94,7 +94,12 @@ function buildYargs(argvInput) {
         alias: "d",
         type: "string",
         description: "a database path, including region",
-        // required: true,
+      },
+      role: {
+        alias: "r",
+        type: "string",
+        description: "a role",
+        default: "admin",
```
echo-bravo-yahoo, line 102: does this apply for every CLI command, or should it be shared to a smaller set of commands?

test/helpers.mjs
```diff
@@ -2,8 +2,8 @@ import { join } from "node:path";
 
 // small helper for sinon to wrap your return value
 // in the shape fetch would return it from the network
-export function f(returnValue) {
-  return { json: async () => returnValue };
+export function f(returnValue, status) {
+  return { json: async () => returnValue, status: status || 200 };
```
echo-bravo-yahoo, line 6: 👍 love it

test/authNZ.mjs
```diff
@@ -0,0 +1,166 @@
...
+      }),
+    };
+  };
+
+  beforeEach(() => {
+    container = setupContainer();
+    container.register({
+      accountClient: awilix.asFunction(mockAccountClient).scoped(),
+      accountCreds: awilix.asClass(AccountKey).scoped(),
+      secretCreds: awilix.asClass(SecretKey).scoped(),
```
echo-bravo-yahoo, line 37: if these are good defaults for all/new tests (and they look to me like they are?), you should move these registrations into `./config/setup-test-container.mjs`.

test/authNZ.mjs
```diff
@@ -0,0 +1,166 @@
...
+    const exit = scope.resolve("exit");
+    const accountCreds = scope.resolve("accountCreds");
+
+    accountCreds.get = stub().throws(new InvalidCredsError());
+
+    await authNZMiddleware(argv);
+    expect(logger.stderr.args[0][0]).to.include("not signed in or has expired");
+    expect(logger.stdout.args[0][0]).to.include(
+      "To sign in, run:\n\nfauna login --profile test-profile\n",
+    );
```
echo-bravo-yahoo, line 63: you can get this effect from this (more readable) syntax. sinon.match will do a substring search.

```
expect(logger.stderr).to.have.been.calledWith(sinon.match("not signed in or has expired"))
```

however, if you know what all of the output should look like, i added `scope.resolve("stdoutStream")` recently, which you can use like [this](https://github.com/fauna/fauna-shell/blob/v3/test/shell.mjs#L123) to make assertions against the whole output.

test/authNZ.mjs
```diff
@@ -0,0 +1,166 @@
...
+  it("should refresh session if account key is invalid", async function () {
+    const argv = { authRequired: true, profile: "test-profile" };
+    const scope = container.createScope();
+
+    await run("db list", scope);
+    const accountCreds = scope.resolve("accountCreds");
+    accountCreds.save = stub();
+
+    const fs = scope.resolve("fs");
+    fs.readFileSync.withArgs(sinon.match(/secret_keys/)).returns("{}");
```
echo-bravo-yahoo, line 76: +1, love the usage of withArgs and the matcher.

test/authNZ.mjs
```diff
@@ -0,0 +1,166 @@
...
+      .withArgs(sinon.match(/access_keys/))
+      .returns(validAccessKeyFile);
+
+    const accountClient = scope.resolve("accountClient");
+    accountClient.whoAmI.onFirstCall().throws(new InvalidCredsError());
+
+    await authNZMiddleware(argv);
+    expect(accountClient.refreshSession.calledOnce).to.be.true;
+    expect(accountCreds.save.calledOnce).to.be.true;
+    expect(accountCreds.save.args[0][0].creds).to.deep.equal({
```
echo-bravo-yahoo, line 87: avoid the `args[0][0]` pattern when it's otherwise expressable.

echo-bravo-yahoo (changes requested):

src/cli.mjs
mwilde345: everything that will be using the middleware, i.e. everything that needs a db key.

echo-bravo-yahoo (approved):

PR #420, opened 2024-11-21
Suppress node warnings

echo-bravo-yahoo:
Ticket(s): FE-6079

### fix packaging, prevent CLI from surfacing nodeJS warnings

the packaging config wasn't right (our npm package should include /dist, not /src, and the binary should be ./dist/cli.cjs, not ./src/user-entrypoint.mjs). this commit changes that.

additionally, nodeJS emits warnings for a variety of things (deprecation, dangling open resources, unusual resource usage). these look unprofessional in the middle of CLI runs. this change disables node's built-in warning logging in production builds only.

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -64,6 +64,32 @@ function buildYargs(argvInput) {
   // https://github.com/yargs/yargs/blob/main/docs/typescript.md?plain=1#L124
   const yargsInstance = yargs(argvInput);
 
+  // these debug commands are used by the tests in environments where they can't mock out the command handler
+  if (
+    process.env.NODE_ENV !== "production" ||
+    process.env.DEBUG_COMMANDS === "true"
+  ) {
+    yargsInstance
+      .command("throw", false, {
+        handler: () => {
+          throw new Error("this is a test error");
+        },
+        builder: {},
+      })
+      .command("reject", false, {
+        handler: async () => {
+          throw new Error("this is a rejected promise");
+        },
+        builder: {},
+      })
+      .command("warn", false, {
+        handler: async () => {
+          process.emitWarning("this is a warning emited on the node process");
+        },
+        builder: {},
+      });
+  }
+
   return yargsInstance
     .scriptName("fauna")
     .middleware([checkForUpdates, logArgv], true)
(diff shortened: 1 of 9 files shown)


src/cli.mjs
```diff
@@ -83,6 +86,12 @@ function buildYargs(argvInput) {
       },
       builder: {},
     })
+    .command("warn", false, {
+      handler: async () => {
+        process.emitWarning("this is a warning emited on the node process");
+      },
+      builder: {},
+    })
```
cleve-fauna, line 94: does this expose the command to the users?

src/cli.mjs
```diff
@@ -27,6 +27,9 @@ export async function run(argvInput, _container) {
   container = _container;
   const logger = container.resolve("logger");
   const parseYargs = container.resolve("parseYargs");
+  if (process.env.NODE_ENV === "production") {
+    process.removeAllListeners("warning");
```
cleve-fauna, line 31: will this have any side effects?

cleve-fauna (commented):
Questions

src/cli.mjs
echo-bravo-yahoo: it's technically runnable, but it doesn't show up in completions, help text, etc. the second arg (false) tells yargs to hide the command.

src/cli.mjs
echo-bravo-yahoo: only if we accidentally register warning events listeners before this. some more context [here](https://nodejs.org/api/process.html#event-warning).

src/cli.mjs
cleve-fauna: Instead of hidden commands in prod, could we use env vars to produce these in dev only.

And then have another flag that toggles warning emission when the NODE_ENV is not production.

src/cli.mjs
echo-bravo-yahoo: we absolutely can do that in the future - we'd just remove all listeners, then add our listener. listeners are additive, so we can't really override so much as remove, then add.

src/cli.mjs
echo-bravo-yahoo: they can be added conditionally behind ENV vars - that's a great idea. these commands are primarily for the tests to exercise certain failures in a controllable ay.

this change _does_ set warning to only emit when NODE_ENV is production.

src/cli.mjs
cleve-fauna: sounds good.

src/cli.mjs
cleve-fauna: cool. mind making that change?

echo-bravo-yahoo: done!

henryfauna (approved):

PR #421, opened 2024-11-21
update dependencies

echo-bravo-yahoo:
resolves 2 medium security vulnerabilities.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "yargs": "^17.7.2"
       },
       "bin": {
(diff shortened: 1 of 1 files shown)


cleve-fauna (commented):
should there be a package.json change?

echo-bravo-yahoo:
> should there be a package.json change?

nope - just needed to update the lock versions. in CI we build from _exactly_ the versions specified in the lock file, so even though there are compatible (by semver/package.json description) dependencies without open security vulns, we would pull the old ones with vulns.

cleve-fauna (approved):

PR #422, opened 2024-11-22
Config

echo-bravo-yahoo:
### brief

this change implements the `--config` flag, adds config parsing to the CLI, and adds env var parsing to the CLI.

it will accept any flags that any commands or subcommands define, but attempts to use a command with an incompatible flag will fail because of strict mode. because of this, it's wisest to only use globally defined flags for the default profile.

### how does config actually work?

given a config file with this content (note that JSON is also valid):
```
default:
  secret: very-secret
  color: true

ci:
  secret: even-more-secret
  color: false
```

and this env var:
```
FAUNA_CONFIG=/path/to/config.yaml
FAUNA_SECRET=not-so-secret
```

running `fauna eval --query "Database.all()"` would:
1. use the env var to locate and read the config file
2. overlay the config file's default profile (since none was specified in flags or env vars) with the env vars and flags, following this precedence: flags > env vars > config entries
3. run the command with the resulting argv. in this case, it would be equivalent to: `fauna eval --secret "not-so-secret" --color --query "Database.all()"`


henryfauna (approved):

mwilde345 (approved):

src/commands/eval.mjs
```diff
@@ -255,6 +255,7 @@ function buildEvalCommand(yargs) {
       query: {
         type: "string",
         description: "the query to run",
+        required: true,
```
jrodewig, line 258: Is this right? I thought a user could specify `--query` or `--file`, and they're mutually exclusive.

PR #423, opened 2024-11-22
add test for incorrect flags

echo-bravo-yahoo:
incorrect flags should error, but there's no tests documenting that.


ecooper (approved):

test/general-cli.mjs
```diff
@@ -39,6 +39,22 @@ describe("cli operations", function () {
     expect(container.resolve("parseYargs")).to.have.been.calledOnce;
   });
 
+  it("should exit with a helpful message if a non-existant flag is not provided", async function () {
```
cleve-fauna, line 42:
```suggestion
  it("should exit with a helpful message if a non-existent flag is provided", async function () {
```

cleve-fauna (approved):

test/general-cli.mjs
echo-bravo-yahoo: thanks for the catch @_@

PR #431, opened 2024-11-22
FE-6024 shell history

ptpaterson:
Ticket(s): FE-6024

## Problem
We use the built-in Node.js REPL for the `shell` command, but don't leverage the native history feature.

## Solution
Setup and register `~/.fauna/history` file and tell the REPL to use it for history.

Add a shell `.clearhistory` shell command to clear the history.

Move file utils to use the containerized `fs` module.

ignore `test/test-homedir`

## Result

https://github.com/user-attachments/assets/89a6736b-dc63-42c8-8c97-79706572e760

## Testing
Updated test container to not stub `node:fs` by default. This required some existing tests to stub it out, but allowed others that need native fs behavior.

Added tests for shell history. Each test specifies a different homedir, so that they tests won't collide with each other.

image


src/lib/file-util.mjs
```diff
@@ -22,15 +20,8 @@ export function fixPath(path) {
...
-    // returns undefined instead of throwing if the file doesn't exist
-    throwIfNoEntry: false,
-  });
-  if (stat === undefined || !stat.isDirectory()) {
-    return false;
-  } else {
-    return true;
-  }
+  const fs = container.resolve("fs");
+  return fs.existsSync(fixPath(path));
```
echo-bravo-yahoo, line 24: this doesn't assert that it's a directory - just that the path exists. rename or change?

test/shell.mjs
```diff
@@ -43,12 +46,33 @@ const v4Object2 = `{
   ]
 }`;
 
+const registerHomedir = (container, subdir = "") => {
+  const __dirname = import.meta.dirname;
+  const homedir = path.join(__dirname, "../test/test-homedir", subdir);
+
+  container.register({
+    homedir: awilix.asFunction(() => homedir).scoped(),
```
echo-bravo-yahoo, line 54: any reason to not register this as a static value? e.g.:
```
  homedir: awilix.asValue(homedir),
```

test/shell.mjs
```diff
@@ -89,6 +113,117 @@ describe("shell", function () {
     it.skip("can read input from a file", async function () {});
 
     it.skip("can set a connection timeout", async function () {});
+
+    const upArrow = "\x1b[A";
+    const downArrow = "\x1b[B";
+
+    it("can keep track of history", async function () {
+      // start the shell
+      const runPromise = run(`shell --secret "secret" --typecheck`, container);
```
echo-bravo-yahoo: i think we should wrap these in a describe block - this lets you group them, and lets you register a beforeEach hook that can unstub `fs`.

if you look at a test run report, you can see how the spec names are built from describe / it block traversals:
image

given that, i think these tests should live somewhere like:
```
shell > common > history > works in one session
... > can be cleared
... > is saved between sessions
```

ideally, the test path builds a readable sentence.

src/config/setup-test-container.mjs
```diff
@@ -51,7 +51,10 @@ export function setupTestContainer() {
     // wrap it in a spy so we can record calls, but use the
     // real implementation
     parseYargs: awilix.asValue(spy(parseYargs)),
-    fs: awilix.asValue(stub(fs)),
+    // Stubbing node:fs globally breaks tests where we want fs to work normally.
+    // Let tests decide how fs should be stubbed out if they don't want the real
+    // implementation.
+    fs: awilix.asValue(fs),
```
echo-bravo-yahoo, line 57: i actually think we should invert this - you can re-register a dependency to overwrite it, so tests that need the real fs implementation can clobber the stub like:
```
// in test
before(() => {
  container.register({ fs: awilix.asValue(fs) });
});
```

most tests should mock fs - they'll be faster and more resilient that way. shell history is an outlier because we don't control the code that does the writing - but in general, we should prefer to mock (and making mocking the default behavior makes it more likely that later devs will choose that).

ptpaterson: I'll give it a shot.

test/shell.mjs
ptpaterson: Right. I feel better about donig this globally now, so I'll go do that.

test/shell.mjs
ptpaterson: 😁

PR #432, opened 2024-11-22
Improve credential handling

mwilde345:
Ticket(s): FE-6152

Probably most helpful to look in order:
* `auth/credentials.mjs` (middleware that runs buildCredentials)
* `auth/accountKeys.mjs` - Makes the `AccountKeyStore` and adds some methods for refreshing account keys
*  `auth/databaseKeys.mjs` - Makes `DatabaseKeyStore` and has refresh methods
*  Other changes are ripping out old `authNZ` stuff and using the new `container.resolve("credentials")`

## Problem
* Existing auth middleware does too much, including a preflight api call to fauna and account api.

## Solution
* Similar to dashboard short-lived keys, make them truly just-in-time by handling 401s with a single refresh and retry attempt.
* Build a singleton `Credentials` class that gets injected.
  * Tests will continue to mock filesystem and network requests, so shouldn't need any more injectables.
* `Credentials` class builds `AccountKeys` and `DatabaseKeys` classes.
  * These provide methods to interact with their respective local credential files (`AccountKeyStore`, `DatabaseKeyStore`), and also provide helpers to get and refresh keys when necessary

* Wrap up `makeAccountRequest` and `client.query` for the account api and fauna api, respectively
  * Check for 401s and refresh or prompt login

## Result
* DB keys and account keys are refreshed behind the scenes and not before every single command
* The correct secret and account key are used, considering the various source of input
* Simple credential argument verification (no `database` and `secret` arg together)
* Commands don't have to worry about creds, they just init the client they need. If they want to interact with credentials, they can.

Example:

```javascript
async function listDatabases(argv) {

  // query the account api
  const accountClient = new FaunaAccountClient();
  const databases = await accountClient.listDatabases();
  logger.stdout(databases);

  // query the fauna api
  const dbClient = await container.resolve("getSimpleClient")(argv);
  const result = await performQuery(dbClient, "Database.all()", undefined, {
    ...argv,
    format: "json",
  });
  logger.stdout(result);

  // see what credentials are being used
  const credentials = container.resolve("credentials");
  logger.debug(
    `
    Account Key: ${credentials.accountKeys.key}\n
    Database Key: ${credentials.databaseKeys.key}`,
    "creds",
  );
}
```

## Testing

Need to update `login` test and add new `credentials` test that verifies middleware sets things up properly

* Manual testing for now until I refactor the old `authNZ` test file.
* Did debug logging to verify that account keys and db keys are refreshed, saved, cleaned up correctly. 
* Verified we don't make duplicate credential classes


src/config/setup-test-container.mjs
```diff
@@ -61,7 +61,7 @@ export function setupTestContainer() {
     getSimpleClient: awilix.asValue(
       stub().returns({ close: () => Promise.resolve() }),
     ),
-    accountClient: awilix.asFunction(stub()),
+    AccountClient: awilix.asValue(stub()),
```
echo-bravo-yahoo, line 64: this'll probably break tests - the code under test will call the stub to construct an accountClient, and then it'll call stub.startOAuthRequest, stub.getToken, and stub.getSession. what about something like:

```suggestion
    AccountClient: awilix.asValue(() => ({ startOAuthRequest: stub(), getToken: stub(), getSession: stub() })),
```

src/lib/command-helpers.mjs
```diff
@@ -1,20 +1,73 @@
...
+  // }
   return headers;
 }
 
 export async function getSimpleClient(argv) {
+  const logger = container.resolve("logger");
+  const { profile, database: path, role, secret } = argv;
+  const accountKey = getAccountKey(profile).accountKey;
+  if (secret) {
+    logger.debug("Using Database secret from command line flag");
```
echo-bravo-yahoo, line 22: the loggers besides stdout and stderr accept a second argument, `component`, which is used to filter verbose output. could you add a component for this (`client` or `network` maybe?). it'll need to be added to these debug calls and to the list of components in `cli.mjs`.

src/config/setup-test-container.mjs
echo-bravo-yahoo: i see the test code using mockAccountClient - if it's a good default mock, maybe we move it to here so the tests don't have to individually register it?

src/commands/database/list.mjs
```diff
@@ -1,19 +1,42 @@
...
-  const accountCreds = container.resolve("accountCreds");
-  const accountKey = accountCreds.get({ key: profile }).accountKey;
-  const databases = await accountClient.listDatabases(accountKey);
+
+  // query the account api
+  const accountClient = new FaunaAccountClient();
+  const databases = await accountClient.listDatabases();
   logger.stdout(databases);
+
+  // query the fauna api
```
henryfauna, line 16: Why do you need to run `Database.all()` here? Is this just leftover from debugging?

src/commands/database/list.mjs
```diff
@@ -1,19 +1,42 @@
...
+    Database Key: ${credentials.databaseKeys.key}`,
+    "creds",
+  );
 }
 
 function buildListCommand(yargs) {
-  return yargs.version(false).help("help", "show help");
+  return yargs
+    .options({
+      ...commonQueryOptions,
```
henryfauna, line 37: These should already be injected at the top level `database` command now.

src/commands/database/list.mjs
mwilde345: yeah just an example usage. i'll get rid of it and update the readme instead

henryfauna (approved):

src/lib/auth/credentials.mjs
```diff
@@ -2,27 +2,50 @@ import { asValue, Lifetime } from "awilix";
...
+    const accountKeys = Object.values(accountKeyData).map(
+      (value) => value.accountKey,
+    );
+    const secretKeyData = this.databaseKeys.keyStore.getFile();
+    Object.keys(secretKeyData).forEach((accountKey) => {
+      if (!accountKeys.includes(accountKey)) {
+        this.databaseKeys.keyStore.updateAccountKey(accountKey);
+        this.databaseKeys.keyStore.deleteAllDBKeysForAccount();
+      }
+    });
```
mwilde345, line 48: Moved the cleanup to this class because
previously I was creating new AccountKeyStorage and DatabaseKeyStorage here which is unnecessary if I just reuse what's already been made.

test/schema/pull.mjs
```diff
@@ -194,7 +194,8 @@ describe("schema pull", function () {
     expect(logger.stdout).to.have.been.calledWith("Change cancelled");
     expect(fs.writeFile).to.have.not.been.called;
     expect(fsp.unlink).to.have.not.been.called;
-    expect(fs.mkdirSync).to.have.not.been.called;
+    // Called twice during Credentials initialization, but not called during the pull command
+    expect(fs.mkdirSync).to.have.been.calledTwice;
```
mwilde345, line 198: Would like to "reset" the stub or somehow separate the `fs` from how it's used in middleware vs in the actual command helper. I imagine it'll be a pain this way to have to keep track of every way that middleware has already interacted with stubs...

.gitignore
```diff
@@ -15,6 +15,7 @@ experiments
 coverage
 test-results.xml
 .history
+.fauna
```
echo-bravo-yahoo, line 18: do the tests create a .fauna dir? if so, can they be mocked? if not, what's .fauna for (in this directory)?

src/lib/auth/credentials.mjs
```diff
@@ -2,27 +2,50 @@ import { asValue, Lifetime } from "awilix";
...
+      "Cannot use both the '--secret' and '--database' options together. Please specify only one.",
+    );
+  } else if (argv.role && argv.secret) {
+    // The '--role' option is not supported when using a secret. Secrets have an
+    // implicit role.
+    logger.warn(
+      "The '--role' option is not supported when using a secret. It will be ignored.",
     );
   }
 };
```
echo-bravo-yahoo, line 21: thoughts on throwing here instead?

test/schema/pull.mjs
echo-bravo-yahoo: yeah - this is an early test and i think the better pattern is to assert that the call you care about happened (`expect(...).to.have.been.calledWith(...)`). this case is a little funky because we want to assert the opposite (not called with...).

test/shell.mjs
```diff
@@ -98,7 +98,8 @@ describe("shell", function () {
 
       // start the shell
       const runPromise = run(`shell --secret "secret"`, container);
-
+      // Wait for the shell to start (print ">")
+      await stdout.waitForWritten();
```
echo-bravo-yahoo, line 102: seems fine.

.gitignore
mwilde345: yes that's from the tests, https://github.com/fauna/fauna-shell/blob/c35346e2bbda5ef48c44a677d7713bb9a385f8bd/src/lib/file-util.mjs#L114

I'll mock that and get rid of this ignore

src/lib/auth/credentials.mjs
mwilde345: Fair! no objection

ptpaterson: +1 for failing outright.

.fauna/credentials/access_keys
```diff
@@ -0,0 +1 @@
+{}
```
ecooper, line 1: I'm not sure we actually want/need this in source control. Did they sneak in after .fauna was removed from gitignore?

src/lib/auth/accountKeys.mjs
```diff
@@ -0,0 +1,130 @@
...
+   * Prompt re-authentication and exit the program;
+   */
+  promptLogin() {
+    const exit = container.resolve("exit");
+    this.logger.stderr(
+      `The requested profile ${this.profile || ""} is not signed in or has expired.\nPlease re-authenticate`,
+    );
+    this.logger.stdout(`To sign in, run:\n\nfauna login\n`);
+    exit(1);
+  }
```
ecooper, line 64: Is this the blessed why for exiting when you want to control the messaging?

Asking for a friend. 🤭

ecooper (approved):
I don't think we need this .fauna files in source control unless I'm missing something.

Otherwise, the biggest thing I notice is this is OO heavy, but on the query side we're more functional + composition. I wouldn't block it now, but we should figure out what our bless pattern is here before release. Otherwise, some of the codebase will be very oo and the rest will be composition and it will be a pain later.

.fauna/credentials/access_keys
mwilde345: oh yikes that shouldn't be in there. ran tests with wrong `homedir` mock

src/lib/auth/accountKeys.mjs
mwilde345: You can very well just `throw new Error(message)` to exit and control the messaging. But it goes to `stderr` only. I guess I just wanted to have a blip of `stdout` after? 

Should probably just throw

mwilde345: we want to avoid calling `exit` in random places

PR #433, opened 2024-11-23
improve type inference on things pulled from awilix

echo-bravo-yahoo:
the types for the container were broken; this commit fixes them. as a result, typescript server can properly infer the type of things resolved off the awilix container. unfortunately, it's likely not going to work in all situations - for instance, i've noticed that JSDoc comments are dropped.

before, looking at something registered with asValue:
image

before, looking at something registered with asClass:
image

after, looking at something registered with asValue:
image

after, looking at something registered with asClass:
image

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -3,17 +3,16 @@
 import chalk from "chalk";
 import yargs from "yargs";
 
-import keyCommand from "./commands/key.mjs";
+import databaseCommand from "./commands/database/database.mjs";
 import evalCommand from "./commands/eval.mjs";
-import shellCommand from "./commands/shell.mjs";
+import keyCommand from "./commands/key.mjs";
 import loginCommand from "./commands/login.mjs";
 import schemaCommand from "./commands/schema/schema.mjs";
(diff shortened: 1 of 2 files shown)


src/cli.mjs
```diff
@@ -3,17 +3,16 @@
...
+import keyCommand from "./commands/key.mjs";
 import loginCommand from "./commands/login.mjs";
 import schemaCommand from "./commands/schema/schema.mjs";
-import databaseCommand from "./commands/database/database.mjs";
-
+import shellCommand from "./commands/shell.mjs";
 import { authNZMiddleware } from "./lib/auth/authNZ.mjs";
 import { checkForUpdates, fixPaths, logArgv } from "./lib/middleware.mjs";
 
-/** @typedef {import('awilix').AwilixContainer<import('./config/setup-container.mjs').modifiedInjectables>} cliContainer */
```
echo-bravo-yahoo, line 16: ran linter which reordered the dependencies

mwilde345 (approved):

PR #435, opened 2024-11-23
Create Database Command

henryfauna:
Ticket(s): FE-6125

## Problem
The create database command is stubbed out and does not actually do anything.

## Solution
- Wire it up with the Fauna v10 driver and run `Database.create`.
- Accept `typechecked`, `protected` and `priority` options.
- Set up some common functions to interact with the v10 driver.

## Result
The `database create` command works using a secret. Providing a database name instead of a secret is not yet supported.

## Testing
Added tests to ensure the command executes the correct query.

### DIFF
--- a/src/commands/database/create.mjs
+++ b/src/commands/database/create.mjs
@@ -36,7 +36,6 @@ async function createDatabase(argv) {
 function buildCreateCommand(yargs) {
   return yargs
     .options({
-      ...commonQueryOptions,
       name: {
         type: "string",
         description: "the name of the database to create",
@@ -53,6 +52,7 @@ function buildCreateCommand(yargs) {
         type: "number",
         description: "user-defined priority assigned to the child database",
       },
+      ...commonQueryOptions,
     })
     .demandOption("name")
     .version(false)
--- a/src/commands/database/create.mjs
+++ b/src/commands/database/create.mjs
@@ -2,19 +2,20 @@
 
 import { fql } from "fauna";
 import { container } from "../../cli.mjs";
-import { runV10Query } from "../../lib/fauna.mjs";
 import { commonQueryOptions } from "../../lib/command-helpers.mjs";
 
 async function createDatabase(argv) {
   const logger = container.resolve("logger");
+  const runV10Query = container.resolve("runV10Query");
 
   await runV10Query({
     url: argv.url,
     secret: argv.secret,
     query: fql`Database.create({
       name: ${argv.name},
-      protected: ${argv.protected ?? false},
-      typechecked: ${argv.typechecked ?? false}
+      protected: ${argv.protected ?? null},
+      typechecked: ${argv.typechecked ?? null},
+      priority: ${argv.priority ?? null},
     })`,
   });
 
--- a/src/commands/database/create.mjs
+++ b/src/commands/database/create.mjs
@@ -1,25 +1,35 @@
 //@ts-check
 
-import { fql } from "fauna";
 import { container } from "../../cli.mjs";
+import { fql, FaunaError } from "fauna";
+import { throwForV10Error } from "../../lib/fauna.mjs";
 import { commonQueryOptions } from "../../lib/command-helpers.mjs";
 
 async function createDatabase(argv) {
   const logger = container.resolve("logger");
   const runV10Query = container.resolve("runV10Query");
 
-  await runV10Query({
-    url: argv.url,
-    secret: argv.secret,
-    query: fql`Database.create({
-      name: ${argv.name},
-      protected: ${argv.protected ?? null},
-      typechecked: ${argv.typechecked ?? null},
-      priority: ${argv.priority ?? null},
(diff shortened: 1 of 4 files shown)


src/lib/fauna.mjs
```diff
@@ -0,0 +1,31 @@
...
+  query,
+  url = undefined,
+  secret = undefined,
+  client = undefined,
+  options = {
+    format: "simple",
+    typecheck: false,
+  },
+}) => {
+  let _client = client;
```
echo-bravo-yahoo, line 19: is there any reason to do this indirection? we can just check if `client` is defined and use that name later.

src/lib/fauna.mjs
```diff
@@ -0,0 +1,31 @@
...
+
+  if (!_client && url && secret) {
+    _client = await getV10Client({ url, secret });
+  } else if (!_client) {
+    throw new Error("No client provided and no url and secret provided");
```
echo-bravo-yahoo, line 24: how about pushing the validation all the way down to `getV10Client`? there's validation at the very top of the stack (argv); might make more sense for this validation to live at the very bottom of the stack and let the middle parts assume valid input.

src/lib/fauna.mjs
```diff
@@ -0,0 +1,31 @@
+export const getV10Client = async ({ url, secret }) => {
+  const { Client } = (await import("fauna")).default;
```
echo-bravo-yahoo, line 2: also, i totally lied to you about there not being an existing v10 client usage in the project :(

the eval command has its own v10 query path that's used by it and the shell command:
https://github.com/fauna/fauna-shell/blob/v3/src/commands/eval.mjs#L101

i think it would make the most sense for eval and shell to only have the formatting logic, and for us to have only one implementation of the client/network stack. take a look at `getSimpleClient` and `performQuery` and see what you think?

src/lib/fauna.mjs
henryfauna: I did it this way so that I can easily close the client created internally if one is not provided.

src/lib/fauna.mjs
henryfauna: Sure I can push it to `getV10Client`. What's the downside of doing validation in `runV10Query` though?

src/lib/fauna.mjs
henryfauna: Yeah that v10 path is not using the driver though so that's why I set up these new helpers. I definitely agree about having a single implementation, I tried consolidating things a bit but it was proving to be difficult. Something to sort out next week :)

src/lib/fauna.mjs
henryfauna: I kept a bit of validation in `runV10Query` just to sanity check the args - let me know if you think there's a better way to do this though

src/commands/database/create.mjs
```diff
@@ -1,19 +1,58 @@
...
+      });
+    }
+    throw e;
+  }
 }
 
 function buildCreateCommand(yargs) {
   return yargs
     .options({
+      ...commonQueryOptions,
```
echo-bravo-yahoo, line 39: the order these are registered in determines the order of the helpfile - thoughts on moving the common ones to the end?

test/database/create.mjs
```diff
@@ -0,0 +1,105 @@
...
+    {
+      args: "--name 'testdb' --secret 'secret' --protected",
+      expected: { name: "testdb", secret: "secret", protected: true },
+    },
+    {
+      args: "--name 'testdb' --secret 'secret' --priority 10",
+      expected: { name: "testdb", secret: "secret", priority: 10 },
+    },
+  ].forEach(({ args, expected }) => {
+    it(`calls runV10Query with correct arguments: ${args}`, async () => {
```
echo-bravo-yahoo, line 57: thoughts on structuring this to be describe the test in a more user-facing & sentence-like way?

thinking something like:
```
describe(`calls fauna with the user-specified arguments`, () => {
  ...
  it(`including ${Object.keys(expected).join(', ')}.`, async () => {
  ...
  })
})
```

test/database/create.mjs
```diff
@@ -0,0 +1,105 @@
...
+    },
+    {
+      error: new ServiceError({
+        error: { code: "unauthorized", message: "whatever" },
+      }),
+      expectedMessage:
+        "Authentication failed: Please either log in using 'fauna login' or provide a valid database secret with '--secret'",
+    },
+  ].forEach(({ error, expectedMessage }) => {
+    it(`handles fauna errors: ${error.code}`, async () => {
```
echo-bravo-yahoo, line 88: similarly, maybe:
```
it(`handles ${error.code} errors when calling fauna`, async () => {
  ...
});
```

echo-bravo-yahoo (approved):

test/database/create.mjs
henryfauna: yup can do

PR #436, opened 2024-11-25
Use `homedir` provided by the container

ptpaterson:
Ticket(s): FE-6024

## Problem

Adding history to the shell requires specifying a file. We are already using files under the `~/.fauna/credentials` folder, but the location is hard-coded and not test-friendly.

## Solution

Use the `homedir` container instead of calling `os.homedir`

## Result

AuthNZ works with the containerized `homedir` and sets precedent for shell history.

## Testing

Updated tests to register `homedir`


test/authNZ.mjs
```diff
@@ -28,9 +32,14 @@ describe.skip("authNZMiddleware", function () {
...
 
   beforeEach(() => {
+    const __filename = fileURLToPath(import.meta.url);
+    const __dirname = path.dirname(__filename);
+    const homedir = path.join(__dirname, "../../test/test-homedir");
+
     container = setupContainer();
     container.register({
       accountClient: awilix.asFunction(mockAccountClient).scoped(),
+      homedir: awilix.asFunction(() => homedir).scoped(),
```
echo-bravo-yahoo, line 42: if this is a good general-purpose default, you can move it up to `./src/config/setup-test-container.mjs`.

test/login.mjs
```diff
@@ -60,12 +62,18 @@ describe("login", function () {
       getToken: stub().resolves({ accessToken: "access-token" }),
     };
   };
+
   beforeEach(() => {
+    const __filename = fileURLToPath(import.meta.url);
+    const __dirname = path.dirname(__filename);
+    const homedir = path.join(__dirname, "../../test/test-homedir");
```
echo-bravo-yahoo, line 69: our runtime is new enough we get a slightly nicer bit of syntactic sugar:
```
const __dirname = import.meta.dirname;
```

echo-bravo-yahoo (approved):

test/login.mjs
ptpaterson: nice

PR #437, opened 2024-11-25
[FE-6127] Delete Database

henryfauna:
Ticket(s): FE-6127

## Problem
We do not support the `fauna database delete` command yet.

## Solution
- Wire up this command following the patterns established by `fauna database create`.
- Ensure we handle some of the common error scenarios a user might run into. For example, the database does not exist.

## Result
You can delete a database using `fauna database delete --name <db_name> --secret <parent_db_secret>`

## Out of Scope
Authenticating with the credentials vended by `fauna login`.

## Testing
Added units to ensure we handle args correctly and call fauna/handle errors correctly.

Screenshot 2024-11-25 at 2 20 57 PM

### DIFF
--- a/src/commands/database/delete.mjs
+++ b/src/commands/database/delete.mjs
@@ -1,0 +1,49 @@
+//@ts-check
+
+import { FaunaError, fql } from "fauna";
+import { container } from "../../cli.mjs";
+import { throwForV10Error } from "../../lib/fauna.mjs";
+import { commonQueryOptions } from "../../lib/command-helpers.mjs";
+
+async function deleteDatabase(argv) {
+  const logger = container.resolve("logger");
+  const runV10Query = container.resolve("runV10Query");
+
+  try {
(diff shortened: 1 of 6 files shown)


echo-bravo-yahoo (approved):

src/commands/database/delete.mjs
```diff
@@ -0,0 +1,49 @@
...
+function buildDeleteCommand(yargs) {
+  return yargs
+    .options({
+      name: {
+        type: "string",
+        description: "the name of the database to delete",
+      },
+      ...commonQueryOptions,
+    })
+    .demandOption("name")
```
echo-bravo-yahoo, line 39: might be clearer in the options block for name:
```
.options({
  name: {
    ...
    required: true
  })
...
```

henryfauna: cool will do

mwilde345 (approved):

PR #438, opened 2024-11-26
implement basic path completions

echo-bravo-yahoo:
this change adds:
- the default yargs completions for commands, options, etc.
- custom completions for profiles (given that you have a config specified)
- custom completions for database path (given that you have specified enough information to make a request to fauna)

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -10,6 +10,7 @@ import queryCommand from "./commands/query.mjs";
 import schemaCommand from "./commands/schema/schema.mjs";
 import shellCommand from "./commands/shell.mjs";
 import { buildCredentials } from "./lib/auth/credentials.mjs";
+import { getDbCompletions, getProfileCompletions } from "./lib/completions.mjs";
 import { configParser } from "./lib/config/config.mjs";
 import { handleParseYargsError } from "./lib/errors.mjs";
 import {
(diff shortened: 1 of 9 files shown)


src/commands/database/list.mjs
```diff
@@ -68,11 +57,24 @@ async function listDatabasesWithSecret(argv) {
...
-async function listDatabases(argv) {
+export async function listDatabases(argv) {
+  let databases;
+  if (argv.secret) {
+    databases = await listDatabasesWithSecret(argv);
+  } else {
+    databases = await listDatabasesWithAccountAPI(argv);
+  }
+  return databases;
+}
```
echo-bravo-yahoo, line 68: refactored this so listDatabases is a generic API layer, and the rendering is done in doListDatabases, which is the command handler

test/helpers.mjs
```diff
@@ -208,3 +208,18 @@ export const mockSecretKeysFile = ({
...
+ */
+export async function eventually(evaluator, ms = 50) {
+  try {
+    return evaluator();
+  } catch (e) {
+    if (ms <= 0) throw e;
+    await new Promise((resolve) => setTimeout(resolve, 1)); // eslint-disable-line no-promise-executor-return
+    return eventually(evaluator, ms - 1);
+  }
+}
```
echo-bravo-yahoo, line 225: added this helper because _deep sigh_ yargs' parse command resolves the promise it returns when it's done parsing... which almost always happens before it's done evaluating async completions. so we can't rely on awaiting the parse promise like we do everywhere else.

src/cli.mjs
```diff
@@ -112,6 +113,50 @@ function buildYargs(argvInput) {
...
+          currentWordFlag === "database" ||
+          currentWord === "-d" ||
+          currentWordFlag === "d"
+        ) {
+          done(await getDbCompletions(currentWord, argv));
+        } else {
+          defaultCompletions();
+        }
+      },
+    )
```
echo-bravo-yahoo, line 159: this is the meat of the change: yargs calls this method when zsh/bash calls `fauna --get-yargs-completions ...`, and then we do some switching based on yargs argv and process.argv and call methods that either sync or async return a list of strings.

we have to use process.argv as well because yargs argv doesn't include the order of the tokens, so we can't easily find `currentWordFlag`...

src/lib/completions.mjs
```diff
@@ -0,0 +1,46 @@
...
+  function getRegionGroup(currentWord) {
+    const rg = regionGroups.filter((rg) => currentWord.startsWith(rg));
+    return rg.length ? rg[0] : undefined;
+  }
+
+  if (!getRegionGroup(currentWord)) {
+    return regionGroups;
+  } else {
+    const { pageSize } = argv;
+    buildCredentials({ ...argv, user: "default" });
```
echo-bravo-yahoo, line 27: this is a glorious, if silly, hack for when we need credentials for making network calls (usually before the credentials middleware runs): just import and run the credential middleware.

src/lib/completions.mjs
```diff
@@ -0,0 +1,45 @@
...
+    const accountClient = new FaunaAccountClient();
+    try {
```
ecooper, line 29: Could we get some docstrings detailing the order of operations here for future us. Looks like we query exactly path, and one directory down if that doesn't work?

src/lib/completions.mjs
```diff
@@ -0,0 +1,45 @@
...
+
+  function getRegionGroup(currentWord) {
+    const rg = regionGroups.filter((rg) => currentWord.startsWith(rg));
+    return rg.length ? rg[0] : undefined;
+  }
+
+  if (!getRegionGroup(currentWord)) {
+    return regionGroups;
+  } else {
+    const { pageSize } = argv;
```
ecooper, line 26: We probably need to debounce this code path. I pushed tab 5 times waiting for a response...which means I probably made >= 5 requests for the same exact data.

ecooper (approved):

cleve-fauna (approved):

PR #443, opened 2024-11-26
Rework eval and shell to fully implement eval

ecooper:
## Problem

The current implementation of eval was not fully ported over, and recent CLI changes have started to add some drift in how we interact with Fauna.

## Solution

This is a "big" PR that's really pretty simple:

* Renames the command to `query` but enables an `eval` alias for the old-timers
* Moves all query helpers out of the query command and into `fauna.mjs` and `faunadb.mjs`. Each module implements the following:
  * `stringExpressionToQuery`: Transform a string into a query
  * `getVXClient`: to return a client
  * `runVXQuery`: execute a query
  * `runVXQueryFromString`: helper to run a query directly from a string input
  * `formatVXError`: support error formatting in a single place
  * `formatVXQueryResponse`: support driver response formatting in a single place
* Use fauna-client for a single entry point to make queries:
  * `runQueryFromString`: wraps the version specific implementations
  * `formatError`: wraps the version specific implementations
  * `formatQueryResponse`: wraps the version specific implementations
* The following flags have been added:
  * `extra`: returns the full API response instead of just the data. This flag is surfaced in the shell via `.toggleExtra`
  * `quiet`: mutes stderr 
  *  `apiVersion`: version has been renamed to not conflict with cli version

## Result

```zsh
  fauna query "Collection.all()"                          run the query and write to stdout
  fauna query -i /path/to/queries.fql                     run the query from a file
  echo "1 + 1" | fauna query -                            run the query from stdin
  fauna query -i /path/to/queries.fql -o /tmp/result.jso  run the query and write to a file
  n
  fauna query -i /path/to/queries.fql -o /tmp/result.jso  run the query and write full API response to a fil
  n --extra            
```

## Testing


src/commands/query.mjs
```diff
@@ -0,0 +1,121 @@
...
+  // resolve the input
+  const expression = resolveInput(argv);
+
+  // get the query handler and run the query
+  try {
+    const results = await container.resolve("runQueryFromString")(expression, argv);
+    const output = formatQueryResponse(results, argv);
+
+    if (argv.output) {
+     container.resolve("fs").writeFileSync(argv.output, output);
```
echo-bravo-yahoo, line 66: do we want to output the result even when we write to file? sort of like `tee`.

src/commands/query.mjs
```diff
@@ -0,0 +1,121 @@
...
+  }
+}
+
+function buildQueryCommand(yargs) {
+  return yargs
+    .positional("query", {
+      type: "string",
+      description: "the query to run; use - to read from stdin",
+    })
+    .nargs('query', 1)
```
echo-bravo-yahoo, line 84: this'll probably cause issues with queries that have spaces in them. is the behavior better if we don't specify a number and let yargs infer? or do we want to enforce string quoting?

see queries like:
```
fauna query let x = 5 --no-color
```

src/commands/shell.mjs
```diff
@@ -72,21 +83,16 @@ async function buildCustomEval(argv) {
 
       let res;
       try {
-        res = await performQuery(client, cmd, undefined, {
+        res = await runQueryFromString(cmd, {
           ...argv,
```
echo-bravo-yahoo, line 87: any reason to destructure, then rest argv instead of passing it as-is?

ecooper: No. This will be cleaned up in a future revision.

src/commands/query.mjs
ecooper: Well, trying to write FQL in zsh without quotes is super dangerous...so I might not do anything about this until we get more feedback from folks.

src/commands/query.mjs
```diff
@@ -0,0 +1,135 @@
...
+        default: false,
+      },
+      ...commonConfigurableQueryOptions,
+    })
+    .example([
+      ['$0 query "Collection.all()" --database us-std/example --role admin', "run the query and write to stdout "],
+      ["$0 query -i /path/to/queries.fql --database us-std/example --role admin", "run the query from a file"],
+      ['echo "1 + 1" | $0 query - --database us-std/example --role admin', "run the query from stdin"],
+      ['$0 query -i /path/to/queries.fql -o /tmp/result.json --database us-std/example --role admin', "run the query and write to a file"],
+      ['$0 query -i /path/to/queries.fql -o /tmp/result.json --extra --database us-std/example --role admin', "run the query and write full API response to a file"],
```
mwilde345, line 123: --role will default to admin when minting a key for the specified database path. Maybe examples should show you don't have to provide role arg.

Also reminds me to go test out custom user roles...

src/lib/fauna-client.mjs
```diff
@@ -90,3 +70,88 @@ export default class FaunaClient {
...
+  try {
+    return await fn(initialSecret);
+  } catch (err) {
+    // If it's a 401, we need to refresh the secret. Let's just do type narrowing here
+    // vs doing another v4 vs v10 check.
+    if (err && (err.httpStatus === 401 || err.requestResult?.statusCode === 401)) {
+      const credentials = container.resolve("credentials");
+
+      await credentials.databaseKeys.onInvalidCreds();
+      const refreshedSecret = await credentials.databaseKeys.getOrRefreshKey();
```
mwilde345, line 92: Since we already awaited the refresh via onInvalidCreds, safe to just access ".key" here

src/lib/faunadb.mjs
```diff
@@ -0,0 +1,158 @@
...
+  return runInContext(expression, createContext(faunadb.query));
+}
+
+const validateQueryParams = ({ query, client, url, secret }) => {
+  if (!query) {
+    throw new Error("A query is required.");
+  } else if (!client && (!url || !secret)) {
+    throw new Error("A client or url and secret are required.");
+  }
+};
```
mwilde345, line 50: Saw some validation of args above somewhere as well. I have some arg validation in Credentials class too. 

Thoughts on validating all args and their combinations in one place in middleware before we get to handlers?

mwilde345 (approved):

henryfauna (approved):

src/lib/fauna-client.mjs
ecooper: I'll clean this up in a follow-up.

src/lib/faunadb.mjs
ecooper: In theory, this check doesn't need to be here at all. The fauna client will barf if we're missing things. This error isn't adding a lot of value as-is.

PR #446, opened 2024-12-02
fix help text not displaying

echo-bravo-yahoo:


src/lib/logger.mjs
```diff
@@ -34,27 +34,15 @@ export function log({
...
-        verboseComponent: {
-          type: "array",
-          default: [],
-        },
-        verbosity: {
-          type: "number",
-          default: 0,
-        },
-      })
-      .version(false).argv;
```
echo-bravo-yahoo, line 52: the config parser runs well before the rest of yargs. as far as i can tell, it looks something like:
1. start parsing argv
2. run config parser
3. check if it's a --help command
4. run before validation middleware
5. run validation
6. run after validation middleware
7. finished parsing argv
8. run command handler

anyway, this block was previously only hit on errors, and exists to pull the populate verbosity so we can decide if we should log the full error stack. now, though, log is called by config parser - which also happens before argv are available. for whatever reason, calling `yargs...argv` is sticky and clobbers the help it would otherwise generate.

so, in the error case, we can instead read the argv out of the parsed yargs object, and do nothing in the config parser case.

ptpaterson (approved):
FWIW, `undefined` is never > nor < nor = any number. So I think you could do `argv?.verbosity >= verbosity`, but I absolutely hate truthiness, so I like it this way instead 😄.

ecooper (approved):

cleve-fauna (approved):

PR #447, opened 2024-12-02
Support running commands against container.

cleve-fauna:
Ticket(s): FE-5990

## Problem

Customers need to easily run CLI commands against a Fauna container. This is useful when they run tests, configure developer environments, etc.

## Solution

Long term we'll do this by:

- adding a local flag to indicate running against a local fauna
- adapting `url` based on presence of local flag. URL is set to localhost:8443 if no`url` is passed in.
- setting secret to 'secret' if no explicit secret is passed in
- using the Fauna API instead of frontdoor for appropriate commands
- fast suceeding logins and other account API specific commands

Start work to support this by:

- adding a local flag to indicate running against a local fauna
- adapting `url` based on presence of local flag. URL is set to localhost:8443 if no`url` is passed in.

 I made this happen by replacing the usage of the object spread syntax with a function call when setting up common arguments. i.e.

```
yargs.options({
...commonQueryOptions,
...otherOptions
})
```

is now

```
yargsWithCommonQueryOptions(yargs).options(otherOptions)
```

This allows us to do the URL adapting.

I will do this over several PRs. This PR is very bare bones; it does the above and nothing else. Over the next few PRs I will also add:

- setting secret to 'secret' if no explicit secret is passed in
- using the Fauna API instead of frontdoor for appropriate commands
- fast suceeding logins and other account API specific commands
## Result

The big change here is that I added a new way to inject the "common options". We need to do this by calling a function now instead of using the object spread syntax when building yargs.

## Testing

Added tests to confirm URL setting works as expected.

Ran the existing tests. My goal with this PR is to ship this new means of yargs construction so this work does not drift from v3.


mwilde345 (approved):

src/lib/command-helpers.mjs
```diff
@@ -66,15 +98,99 @@ export const commonConfigurableQueryOptions = {
...
+        );
+        await credentials.databaseKeys.onInvalidCreds();
+        const updatedArgs = await queryArgs(args);
+        return await originalQuery(...updatedArgs);
+      }
+      return result;
+    });
+  };
+
+  return client;
```
ptpaterson, line 138: Do we need to return a monkey-patched version of the client which could be v4 or v10? Or can we return a new POJO with a `query` property?

cleve-fauna: this is a rebasing error. this method is not used. i will remove it.

src/lib/command-helpers.mjs
```diff
@@ -66,15 +98,99 @@ export const commonConfigurableQueryOptions = {
...
- * Validate that the user has specified either a database or a secret.
- * This check is not required for commands that can operate at a
- * "root" level.
- * @param {object} argv 
- * @param {string} argv.database - The database to use
- * @param {string} argv.secret - The secret to use
+ * This function will return a v4 or v10 client based on the version provided in the argv.
+ *   onInvalidCreds decides whether or not we retry or ask the user to re-enter their secret.
+ * @param {*} argv
+ * @returns { Promise<any> } - A Fauna client
```
ptpaterson: I miss real real typescript 😅. But, seriously, how about this?
```suggestion
 * @returns { Promise<{ query: (queryValue: any, queryOptions?: any) => Promise<any>}> } - A Fauna client
```

ptpaterson: nvm I am caught up on the other comments.

ptpaterson (approved):

PR #448, opened 2024-12-03
Finish the `database list` command

henryfauna:
Ticket(s): FE-6126

## Problem
The `database list` command does not support `--secret` or `--database`.

## Solution
Support the `--secret` and `--database` flags by using the account API if no secret is given.

## Result
The `database list` command supports all expected flags.

For example, the following commands should now all work:
1. `fauna database list`
2. `fauna database list --secret 'my-secret'`
3. `fauna database list --database 'us-std/my-db'`
4. `fauna database list --json`
5. `fauna database list --pageSize 10`

## Testing
Screenshot 2024-12-02 at 4 22 06 PM

### DIFF
--- a/src/commands/database/list.mjs
+++ b/src/commands/database/list.mjs
@@ -1,6 +1,7 @@
 //@ts-check
 
 import { FaunaError } from "fauna";
+
 import { container } from "../../cli.mjs";
 import { throwForError } from "../../lib/fauna.mjs";
 import { yargsWithCommonQueryOptions } from "../../lib/command-helpers.mjs";
@@ -56,9 +57,9 @@ async function listDatabasesWithSecret(argv) {
 
 async function listDatabases(argv) {
   if (argv.secret) {
-    return listDatabasesWithSecret(argv);
+    listDatabasesWithSecret(argv);
   } else {
-    await listDatabasesWithAccountAPI(argv);
+    listDatabasesWithAccountAPI(argv);
   }
 }
 
--- a/src/commands/database/list.mjs
+++ b/src/commands/database/list.mjs
@@ -7,11 +7,20 @@ import { throwForError } from "../../lib/fauna.mjs";
 import { yargsWithCommonQueryOptions } from "../../lib/command-helpers.mjs";
 import { FaunaAccountClient } from "../../lib/fauna-account-client.mjs";
 
-const OUTPUT_FIELDS = ["name"];
+// Narrow the output fields based on the provided flags.
+const getOutputFields = (argv) => {
+  if (!argv.secret && !argv.database) {
+    // If we are listing top level databases the region group
+    // needs to be included as database names can be re-used across
+    // regions.
+    return ["name", "region_group"];
+  }
+  return ["name"];
+};
 
-function pickOutputFields(databases) {
+function pickOutputFields(databases, argv) {
   return databases.map((d) =>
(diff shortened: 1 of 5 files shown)


src/commands/database/list.mjs
```diff
@@ -1,33 +1,98 @@
...
-  const logger = container.resolve("logger");
+// Narrow the output fields based on the provided flags.
+const getOutputFields = (argv) => {
+  if (!argv.secret && !argv.database) {
+    // If we are listing top level databases the region group
+    // needs to be included as database names can be re-used across
+    // regions.
+    return ["name", "region_group"];
+  }
+  return ["name"];
```
cleve-fauna, line 17: why are we choosing not to include has `has_children`, `region_group` and `path`?

I realize that the core APIs can't give us this at this time.

Is it strictly for that reason?

src/commands/database/list.mjs
```diff
@@ -1,33 +1,98 @@
...
+  return databases.map((d) =>
+    getOutputFields(argv).reduce((acc, field) => {
+      acc[field] = d[field];
+      return acc;
+    }, {}),
   );
 }
 
+async function listDatabasesWithAccountAPI(argv) {
+  const { pageSize, database, json } = argv;
```
cleve-fauna, line 30: why are we not supporting pagination?

src/commands/database/list.mjs
```diff
@@ -1,33 +1,98 @@
...
+    .help("help", "show help")
+    .example([
+      ["$0 database list", "list all databases"],
+      [
+        "$0 database list --database 'us-std/example'",
+        "list all databases under us-std/example",
+      ],
+      [
+        "$0 database list --secret 'my-secret'",
+        "list all databases using the provided database secret",
```
cleve-fauna, line 91: what does `fauna database list --secret 'my-secret' --database 'eu-std/example'` do?

test/database/list.mjs
```diff
@@ -0,0 +1,77 @@
+//@ts-check
+
+import { expect } from "chai";
+import { ServiceError } from "fauna";
+import sinon from "sinon";
+
+import { run } from "../../src/cli.mjs";
+import { setupTestContainer as setupContainer } from "../../src/config/setup-test-container.mjs";
+
+describe("database list", () => {
```
cleve-fauna, line 10: what's the plan to test we get the output we actually we want?

Is that being saved for the integration test story?

cleve-fauna (commented):
Questions

src/commands/database/list.mjs
henryfauna: Yeah I've included `region_group` if listing at the top level as that is needed to disambiguate databases with the same name across region groups. Otherwise I've just kept it to name to be consistent when listing from both APIs. Theoretically you also know the region and path if listing with the `--database` flag.

src/commands/database/list.mjs
henryfauna: You get this error message `Cannot use both the '--secret' and '--database' options together. Please specify only one.`

src/commands/database/list.mjs
henryfauna: I didn't want to tackle that here but I can in a follow up

cleve-fauna (approved):

mwilde345 (approved):

PR #449, opened 2024-12-03
FE-5990 Set argv.secret to 'secret' when --local passed in and --secret is not. Always respect argv.secret when getting a secret with FaunaClient.getSecret.

cleve-fauna:
Ticket(s): FE-5990

## Problem

`--local` must set up user to user a local container. This means:

1. Set url to `http://localhost:8443` if no explicit  `--url` is provided.
2. Set secret to `secret` if no explicit `--secret` is provided.
3. Call container, and not the account API, for list databases, create key
4. Fast return account API commands like login.

**(1)** is already shipped.

This PR implements **(2)**.

Additionally, I noticed that we were not respecting `argv.secret` in query or shell. This PR fixes that.

## Solution

I updated the coercion based on local's present to set `argv.secret` if no explicit secret is given.
## Result

Users can easy use a default Fauna container; and can tweak it with a custom secret or URL

## Out of Scope
1. Abnegating the use of the account API when `--local` is passed.
2. Fast-completing account API commands like `login`

## Testing

- Ran the full suite.
- Ran the linter.

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -104,6 +104,18 @@ function buildYargs(argvInput) {
     .env("FAUNA")
     .config("config", configParser)
     .middleware([checkForUpdates, logArgv], true)
+    .middleware((argv) => {
+      if (!argv.url) {
+        if (argv.local) {
+          argv.url = 'http://localhost:8443';
+        } else {
+          argv.url = 'https://db.fauna.com';
+        }
+      }
+      if (!argv.secret && argv.local) {
+        argv.secret = "secret";
+      }
+    })
     .middleware([fixPaths, buildCredentials], false)
     .command(queryCommand)
     .command("shell", "start an interactive shell", shellCommand)
--- a/src/lib/fauna-client.mjs
+++ b/src/lib/fauna-client.mjs
@@ -76,7 +76,7 @@ export default class FaunaClient {
  * fetches a secret for the current credentials.
  * @param opts {Object}
  * @param opts.secretArg {string} The secret passed as an argument, if any.
- * @return {string} the secret
+ * @return {Promise<string>} the secret
  */
 export async function getSecret({ secretArg }) {
   if (secretArg !== undefined) {
--- a/src/lib/fauna-client.mjs
+++ b/src/lib/fauna-client.mjs
@@ -75,7 +75,7 @@ export default class FaunaClient {
  * Gets a secret. Defaults to the secretArg. If the secretArg is undefined
  * fetches a secret for the current credentials.
  * @param opts {Object}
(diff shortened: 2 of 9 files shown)


ptpaterson (approved):

src/lib/fauna-client.mjs
```diff
@@ -71,7 +71,17 @@ export default class FaunaClient {
   }
 }
 
-export const getSecret = async () => {
+/**
+ * Gets a secret. Defaults to the secretArg. If the secretArg is undefined
+ * fetches a secret for the current credentials.
+ * @param opts {Object}
+ * @param opts.secretArg {string} The secret passed as an argument, if any.
```
cleve-fauna, line 78:
```suggestion
 * @param opts.secretArg {string | undefined} The secret passed as an argument, if any.
```

src/lib/fauna-client.mjs
```diff
@@ -71,7 +71,17 @@ export default class FaunaClient {
...
+ * Gets a secret. Defaults to the secretArg. If the secretArg is undefined
+ * fetches a secret for the current credentials.
+ * @param opts {Object}
+ * @param opts.secretArg {string | undefined} The secret passed as an argument, if any.
+ * @return {Promise<string>} the secret
+ */
+export async function getSecret({ secretArg }) {
+  if (secretArg !== undefined) {
+    return secretArg;
+  }
```
mwilde345, line 84: Credential related arguments should be settled in the middleware so we can have all the various sources and defaults resolved in one place. I am thinking here:
https://github.com/fauna/fauna-shell/blob/v3/src/lib/auth/databaseKeys.mjs#L47-L61

have:
```
if (argv.secret) {
    key = argv.secret;
    keySource = "user";
  }  else if (argv.url) {
    key = "secret";
    keySource = "local-default"
 } else
...
```

after that this `getSecret` function should find "secret" in `credentials.databaseKeys.key`

cleve-fauna: OK seems like the middleware has a bug then. I'll fix it.

mwilde345 (approved):

src/cli.mjs
```diff
@@ -104,6 +104,18 @@ function buildYargs(argvInput) {
...
+        if (argv.local) {
+          argv.url = 'http://localhost:8443';
+        } else {
+          argv.url = 'https://db.fauna.com';
+        }
+      }
+      if (!argv.secret && argv.local) {
+        argv.secret = "secret";
+      }
+    })
```
ecooper, line 118: If you move this into a function, we can unit test it as an individual middleware.

cleve-fauna: will do. i love my end to end tests tho.

PR #450, opened 2024-12-03
Move middleware to separate method for testability. Add debug logging when --local applies changes to argv.

cleve-fauna:
Ticket(s): FE-###

## Problem

## Solution

## Result

## Testing

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -11,7 +11,7 @@ import schemaCommand from "./commands/schema/schema.mjs";
 import shellCommand from "./commands/shell.mjs";
 import { buildCredentials } from "./lib/auth/credentials.mjs";
 import { configParser } from "./lib/config/config.mjs";
-import { checkForUpdates, fixPaths, logArgv } from "./lib/middleware.mjs";
+import { applyLocalArg, checkForUpdates, fixPaths, logArgv } from "./lib/middleware.mjs";
 
 /** @typedef {import('awilix').AwilixContainer<import('./config/setup-container.mjs').modifiedInjectables> } cliContainer */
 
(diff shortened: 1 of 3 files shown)


ecooper (approved):
Nice.

src/cli.mjs
```diff
@@ -104,18 +104,7 @@ function buildYargs(argvInput) {
...
-        } else {
-          argv.url = 'https://db.fauna.com';
-        }
-      }
-      if (!argv.secret && argv.local) {
-        argv.secret = "secret";
-      }
-    })
+    .middleware(applyLocalArg)
     .middleware([fixPaths, buildCredentials], false)
```
echo-bravo-yahoo, line 108: add it to the array on line 108? makes order clearer if they're all array-based or all standalone calls. e.g.:
```
.middleware([applyLocalArgs, fixPaths, buildCredentials], false)
```

echo-bravo-yahoo (approved):

src/lib/middleware.mjs
```diff
@@ -44,3 +44,29 @@ export function checkForUpdates(argv) {
...
+ * If --local is provided and --secret is not, argv.secret is
+ * set to 'secret'.
+ * @param {import('yargs').Arguments} argv
+ * @returns {void}
+*/
+export function applyLocalArg(argv) {
+  const logger = container.resolve("logger");
+  if (!argv.url) {
+    if (argv.local) {
+      argv.url = 'http://localhost:8443';
```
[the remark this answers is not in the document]
cleve-fauna: i'll make them constants nerds.

src/cli.mjs
cleve-fauna: sure thing.

src/lib/middleware.mjs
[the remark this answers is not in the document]
cleve-fauna: done!

src/cli.mjs
cleve-fauna: done!

PR #453, opened 2024-12-03
FE-6024 Implement shell history

ptpaterson:
Ticket(s): FE-6024

## Problem
We use the built-in Node.js REPL for the `shell` command, but don't leverage the native history feature.

## Solution
Setup and register `~/.fauna/history` file and tell the REPL to use it for history.

Add a shell `.clearhistory` shell command to clear the history.

Move file utils to use the containerized `fs` module.

ignore `test/test-homedir`

## Result

https://github.com/user-attachments/assets/89a6736b-dc63-42c8-8c97-79706572e760

## Testing
Updated test container to not stub `node:fs` by default. This required some existing tests to stub it out, but allowed others that need native fs behavior.

Added tests for shell history. Each test specifies a different homedir, so that they tests won't collide with each other.

image


src/config/setup-test-container.mjs
```diff
@@ -42,10 +42,12 @@ export function setupTestContainer() {
   const container = setupCommonContainer();
 
   const thingsToManuallyMock = automock(container);
-  const customfs = stub(fs);
+  const customfs = stub({ ...fs });
```
mwilde345, line 45: what is this doing?

mwilde345 (approved):

src/commands/shell.mjs
```diff
@@ -1,23 +1,37 @@
...
+  const homedir = container.resolve("homedir")();
+  const historyDir = path.join(homedir, ".fauna");
+  if (!dirExists(historyDir)) {
+    fs.mkdirSync(historyDir, { recursive: true });
+  }
+  const historyFile = path.join(historyDir, "history");
+  if (!fileExists(historyFile)) {
+    fs.writeFileSync(historyFile, "");
+  }
+
```
cleve-fauna, line 34: put the file management code in `src/lib/file-util.mjs` which is where the other files we manage are maintained.

src/commands/shell.mjs
```diff
@@ -29,9 +43,18 @@ async function shellCommand(argv) {
     input: container.resolve("stdinStream"),
     eval: await buildCustomEval(argv),
     terminal: true,
+    historySize: 1000,
```
cleve-fauna, line 46: Is this a valid argument?

I don't see it here: https://nodejs.org/docs/latest-v22.x/api/repl.html#replstartoptions

cleve-fauna (commented):
Please reorganize the code a bit and take a look at the question.

src/commands/shell.mjs
cleve-fauna: doesn't credential setup happen in there?

ptpaterson: doh... CredentialStorage is indeed in there. I was trying to go off of memory and I was wrong.

src/commands/shell.mjs
ptpaterson: I am now struggling to re-find it, but I thought I had a trustworthy example that included it.

I have tested and it WORKS. It seems prudent to leave some sort of limit here...

ah
> NODE_REPL_HISTORY_SIZE: Controls how many lines of history will be persisted if history is available. Must be a positive number. Default: 1000.

Since there is in fact a limit built in, I'll just remove the explicit one.

src/commands/shell.mjs
ptpaterson: I added new functions to file-util to initialize and clear history storage.

PR #454, opened 2024-12-03
[FE-6155] - Support `--database` in database commands

henryfauna:
Ticket(s): FE-6155

## Problem
The `database create` and `database delete` commands do not support specifying a target database using `--database`.

## Solution
Update the commands to support this flag by using the `getSecret` helper. Ensure bad creds are only retried/refreshed if we are not using a user provided secret. Some validation was also included to check that at least one of these flags is provided.

## Result
The `--database` flag is supported by `database create` and `database delete`.

## Testing
Added tests to cover the usage of `--database`.

### DIFF
--- a/src/commands/database/create.mjs
+++ b/src/commands/database/create.mjs
@@ -5,14 +5,11 @@ import { FaunaError } from "fauna";
 import { container } from "../../cli.mjs";
 import { throwForError } from "../../lib/fauna.mjs";
 import { formatObjectForShell } from "../../lib/misc.mjs";
+import { validateSecretOrDatabase } from "./database.mjs";
 import { getSecret, retryInvalidCredsOnce } from "../../lib/fauna-client.mjs";
 
 function validate(argv) {
-  if (!argv.secret && !argv.database) {
-    throw new Error(
-      "No secret or database provided. Please use either --secret or --database.",
-    );
-  }
+  validateSecretOrDatabase(argv);
   return true;
 }
 
--- a/src/commands/database/create.mjs
+++ b/src/commands/database/create.mjs
@@ -37,15 +37,9 @@ async function createDatabase(argv) {
   const logger = container.resolve("logger");
 
   try {
-    if (argv.secret === secret) {
-      // If we are using a user provided secret, we should not
-      // try to refresh it if it is bad.
-      await runCreateQuery(secret, argv);
-    } else {
-      await retryInvalidCredsOnce(secret, async (secret) =>
-        runCreateQuery(secret, argv),
-      );
-    }
+    await retryInvalidCredsOnce(secret, async (secret) =>
+      runCreateQuery(secret, argv),
+    );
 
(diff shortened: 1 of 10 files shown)


src/commands/database/create.mjs
```diff
@@ -63,8 +87,19 @@ function buildCreateCommand(yargs) {
...
       },
     })
+    .check(validate)
     .version(false)
-    .help("help", "show help");
+    .help("help", "show help")
+    .example([
+      [
+        "$0 database create --name 'my-database' --database 'us-std/example'",
+        "Create a database named 'my-database' under `us-std/example`.",
```
echo-bravo-yahoo, line 96: example descriptions! 💟

src/commands/database/delete.mjs
```diff
@@ -4,18 +4,41 @@ import { FaunaError } from "fauna";
...
 
-async function deleteDatabase(argv) {
+function validate(argv) {
+  if (!argv.secret && !argv.database) {
+    throw new Error(
+      "No secret or database provided. Please use either --secret or --database.",
+    );
+  }
+  return true;
+}
```
echo-bravo-yahoo, line 16: could extract to `database/database.mjs` and import it in each subcommand.

src/commands/database/create.mjs
```diff
@@ -5,23 +5,47 @@ import { FaunaError } from "fauna";
...
+  try {
+    if (argv.secret === secret) {
+      // If we are using a user provided secret, we should not
+      // try to refresh it if it is bad.
+      await runCreateQuery(secret, argv);
+    } else {
+      await retryInvalidCredsOnce(secret, async (secret) =>
+        runCreateQuery(secret, argv),
+      );
+    }
```
mwilde345, line 48: My assumption was this behavior is already handled via: https://github.com/fauna/fauna-shell/blob/5d621c949d4841bd6d36349dba9474e2f184ee60/src/lib/auth/databaseKeys.mjs#L76-L80

Did it not work that way in your testing?

henryfauna: Ahh cool I didn't know this was already baked in

src/commands/database/delete.mjs
henryfauna: Yeah good idea

src/commands/database/create.mjs
```diff
@@ -5,23 +5,47 @@ import { FaunaError } from "fauna";
...
+
+  try {
+    if (argv.secret === secret) {
+      // If we are using a user provided secret, we should not
+      // try to refresh it if it is bad.
+      await runCreateQuery(secret, argv);
+    } else {
+      await retryInvalidCredsOnce(secret, async (secret) =>
+        runCreateQuery(secret, argv),
+      );
```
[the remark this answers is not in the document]
henryfauna: Turned out we already make this check elsewhere

mwilde345 (approved):

PR #455, opened 2024-12-03
Bugbash changes

echo-bravo-yahoo:
View changes commit-by-commit.


cleve-fauna:
tests are all broken.

ecooper (approved):

cleve-fauna (approved):
Why don't we just run `prettier` and `lint --fix` on each commit to a PR?

Each commit to local perhaps we could toggle with an env var as well? Noting that I appreciate the desire to avoid getting in contributor's way and thus not forcing it to be run.

echo-bravo-yahoo:
> Why don't we just run `prettier` and `lint --fix` on each commit to a PR?
> 
> Each commit to local perhaps we could toggle with an env var as well? Noting that I appreciate the desire to avoid getting in contributor's way and thus not forcing it to be run.

Git hooks on (local) commit are a truly miserable experience. I've considered (but not implemented) pre-push git hooks or github action tasks. Thoughts on either of those?

PR #456, opened 2024-12-04
Add colorized output to database, query, and shell commands

ecooper:
## Problem

The database, query, and shell commands all output JSON. It's pretty printed, but does not support color. It's hard to scan as a result

## Solution

For the database, query, and shell commands support colorized json output by default. This is done through the `formatObjectForShell` and `formatFullErrorForShell` helpers. The rest of this PR is largely just plumbing the color option through and updating tests.

## Result

Query w/ color:
Screenshot 2024-12-03 at 4 06 25 PM

Color w/out color:
Screenshot 2024-12-03 at 4 06 38 PM

A different, fauna-like theme can be added in a follow-up PR.

## Testing

Tests were updated to handle color changes.

### DIFF
--- a/src/commands/database/delete.mjs
+++ b/src/commands/database/delete.mjs
@@ -16,7 +16,9 @@ async function deleteDatabase(argv) {
       secret: argv.secret,
       query: fql`Database.byName(${argv.name}).delete()`,
     });
-    logger.stdout(`Database '${argv.name}' was successfully deleted.`);
+
+    // We use stderr for messaging and there's no stdout output for a deleted database
+    logger.stderr(`Database '${argv.name}' was successfully deleted.`);
   } catch (e) {
     if (e instanceof FaunaError) {
       throwForError(e, {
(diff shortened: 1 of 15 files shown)


src/commands/database/delete.mjs
```diff
@@ -16,7 +16,9 @@ async function deleteDatabase(argv) {
       secret: argv.secret,
       query: fql`Database.byName(${argv.name}).delete()`,
     });
-    logger.stdout(`Database '${argv.name}' was successfully deleted.`);
+
+    // We use stderr for messaging and there's no stdout output for a deleted database
+    logger.stderr(`Database '${argv.name}' was successfully deleted.`);
```
cleve-fauna, line 21: why?

cleve-fauna (approved):

src/commands/database/delete.mjs
cleve-fauna: thanks. makes sense

henryfauna (approved):

PR #460, opened 2024-12-04
disable strict options parsing

echo-bravo-yahoo:
env variables for query settings (FAUNA_SECRET, etc) are too useful to not
have set, but they fail strict options parsing for commands that don't
use them. this commit turns off strict options parsing for now while we find a
better way to address this.


ecooper (approved):

ptpaterson (approved):
sounds good to me

PR #462, opened 2024-12-04
Improve error handling

ecooper:
## Problem

All errors are thrown and treated the same. This means validation errors, known error states, and unexpected errors get the same treatment.

## Solution

This PR updates our error handling for adding capabilities for three classes of errors:

* Validation errors (ours and yargs): Arguments or options are invalid. Print help, display the validation error.
* Command errors:  Errors we intentionally throw in commands or middleware. Print the error only.
* Unknown errors: Errors that are not validation, yargs, or command errors. So things we haven't thrown. Tell the user it's unexpected, print the error, and display a message for bug reports.

Additionally, for debugging purposes, if `cause` is available on an error that will be outputted as well.

## Result

Validation error/unknown argument (ignore the red background, that's my shell):
Screenshot 2024-12-04 at 10 59 59 AM

Command error/not logged in
Screenshot 2024-12-04 at 11 00 33 AM

Command error/bad query syntax
Screenshot 2024-12-04 at 11 02 16 AM

Unknown error/throw
Screenshot 2024-12-04 at 11 22 08 AM

With verbosity
Screenshot 2024-12-04 at 11 56 43 AM

## Testing

The tests have been updated to reflect this new behavior.

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -1,5 +1,6 @@
 // @ts-check
 
+import { util } from "chai";
 import chalk from "chalk";
 import yargs from "yargs";
 
@@ -60,14 +61,15 @@ export async function run(_argvInput, _container) {
         );
         epilogue = `\n${BUG_REPORT_MESSAGE}`;
 
-        logger.debug(`unknown error thrown: ${e.name}`);
+        logger.debug(`unknown error thrown: ${e.name}`, "error");
+        logger.debug(util.inspect(e, true, 2, false), "error");
       } else {
         // Otherwise, just use the error message
         subMessage = chalk.red(e.message);
       }
     }
 
-    // If the error has a truthy displayHelp property, render the help text. Otherwise, just use the error message.
+    // If the error has a truthy hideHelp property, do not render the help text. Otherwise, just use the error message.
     logger.stderr(
       `${e.hideHelp ? "" : `${chalk.reset(await builtYargs.getHelp())}\n\n`}${subMessage}`,
     );
@@ -78,7 +80,7 @@ export async function run(_argvInput, _container) {
 
     // Log the stack if it exists
     logger.fatal(e.stack, "error");
-    logger.fatal(e.cause?.stack, "cause");
+    logger.fatal(e.cause?.stack, "error");
 
     // If the error has an exitCode property, use that. Otherwise, use 1.
     container.resolve("errorHandler")(
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -10,8 +10,14 @@ import queryCommand from "./commands/query.mjs";
 import schemaCommand from "./commands/schema/schema.mjs";
 import shellCommand from "./commands/shell.mjs";
 import { buildCredentials } from "./lib/auth/credentials.mjs";
+import { isUnknownError } from "./lib/command-helpers.mjs";
 import { configParser } from "./lib/config/config.mjs";
-import { applyLocalArg, checkForUpdates, fixPaths, logArgv } from "./lib/middleware.mjs";
+import {
+  applyLocalArg,
+  checkForUpdates,
+  fixPaths,
+  logArgv,
+} from "./lib/middleware.mjs";
 
 /** @typedef {import('awilix').AwilixContainer<import('./config/setup-container.mjs').modifiedInjectables> } cliContainer */
 
@@ -21,6 +27,8 @@ export let container;
 export let builtYargs;
 
 export let argvInput;
+
+const BUG_REPORT_MESSAGE = `If you believe this is a bug, please report this issue on GitHub: https://github.com/fauna/fauna-shell/issues`;
 
 /**
  * @param {string|string[]} _argvInput - The command string provided by the user or test. Parsed by yargs into an argv object.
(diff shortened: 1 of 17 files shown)


src/cli.mjs
```diff
@@ -42,15 +50,41 @@ export async function run(_argvInput, _container) {
...
     if (argvInput.length > 0) {
-      subMessage = chalk.red(e.message);
+      // If the error isn't one of our known errors, wrap it in a generic error message.
+      if (isUnknownError(e)) {
+        subMessage = chalk.red(
+          `An unexpected error occurred...\n\n${e.message}`,
+        );
+        epilogue = `\n${BUG_REPORT_MESSAGE}`;
+
+        logger.debug(`unknown error thrown: ${e.name}`);
```
ptpaterson, line 63: Can we print the whole error, not just the name? That would include the stack trace and might aid in debugging on our end if a customer reports something.

src/cli.mjs
```diff
@@ -42,15 +50,41 @@ export async function run(_argvInput, _container) {
...
+      } else {
+        // Otherwise, just use the error message
+        subMessage = chalk.red(e.message);
+      }
     }
-    const message = `${chalk.reset(await builtYargs.getHelp())}\n\n${subMessage}`;
-    logger.stderr(message);
+
+    // If the error has a truthy displayHelp property, render the help text. Otherwise, just use the error message.
```
ptpaterson, line 70: if not falsy hideHelp instead?

ptpaterson: or this?
```js
if (!e.hideHelp){
  logger.stderr(`${chalk.reset(await builtYargs.getHelp())}\n\n`}${subMessage}`);
}
```

ecooper: Ah, yeah, I had it inverted but that broke a lot of the current assumptions. I'll update.

src/cli.mjs
ecooper: I'll do another debug message with the output of util.inspect. It will not do much, but it could be helpful for any fauna driver errors that bubble up.

ecooper: The stack is written via the logger output later on, btw.

src/cli.mjs
```diff
@@ -42,15 +51,44 @@ export async function run(_argvInput, _container) {
     let subMessage = chalk.reset(
       "Use 'fauna <command> --help' for more information about a command.",
     );
+    let epilogue = "";
```
echo-bravo-yahoo, line 54: thoughts on extracting this to a helper function and typing the error object? might make it easier to recall/find how to get certain error behaviors.

edit: found the types. still think this might be worth extracting just because our runner/entrypoint is now like 90% error handler.

src/cli.mjs
```diff
@@ -93,7 +131,7 @@ function buildYargs(argvInput) {
       })
       .command("warn", false, {
         handler: async () => {
-          process.emitWarning("this is a warning emited on the node process");
+          process.emitWarning("this is a warning emitted on the node process");
```
echo-bravo-yahoo, line 134: my poor spelling getting all beaten up between you and cleve 😔

src/lib/command-helpers.mjs
```diff
@@ -1,19 +1,22 @@
 //@ts-check
 
-// used for queries customers can't configure that are made on their behalf
+// used for queries customers can't config  ure that are made on their behalf
```
echo-bravo-yahoo, line 3:
```suggestion
// used for queries customers can't configure that are made on their behalf
```

src/lib/command-helpers.mjs
```diff
@@ -54,12 +128,15 @@ const COMMON_QUERY_OPTIONS = {
  * @param {object} argv
  * @param {string} argv.database - The database to use
  * @param {string} argv.secret - The secret to use
+ * @param {boolean} argv.local - Whether to use a local Fauna container
```
echo-bravo-yahoo, line 131: should really have a ticket for typing argv correctly...

src/lib/fauna.mjs
```diff
@@ -73,7 +77,7 @@ export const runQuery = async ({
 }) => {
   // Check for required arguments.
   if (!query) {
-    throw new Error("A query is required.");
+    throw new CommandError("A query is required.");
```
echo-bravo-yahoo, line 80: is this not validation? the line's fuzzy, but it seems to me that validation errors are sort of all errors we can determine strictly from argv without touching fs, network, etc.

echo-bravo-yahoo (approved):

ptpaterson (approved):
I don't know how yet, but I would very much like to print out more info for Fauna Service errors, including error code and other information like the actual abort object for aborts and contraints for constraint errors. At least at debug level. But this looks good and I think tweaks to handling specific errors can be done in a follow up.

src/lib/fauna.mjs
ptpaterson: +1, because hideHelp is false on ValidationError

ecooper:
> I don't know how yet, but I would very much like to print out more info for Fauna Service errors, including error code and other information like the actual abort object for aborts and contraints for constraint errors. At least at debug level. But this looks good and I think tweaks to handling specific errors can be done in a follow up.

We do in query via --extra.

src/cli.mjs
```diff
@@ -42,15 +51,44 @@ export async function run(_argvInput, _container) {
...
-    const message = `${chalk.reset(await builtYargs.getHelp())}\n\n${subMessage}`;
-    logger.stderr(message);
+
+    // Log the stack if it exists
     logger.fatal(e.stack, "error");
-    const exitCode = e.exitCode !== undefined ? e.exitCode : 1;
-    container.resolve("errorHandler")(e, exitCode);
+    if (e.cause) {
+      logger.fatal(e.cause?.stack, "error");
+    }
```
ptpaterson, line 85: don't need null coalesce here
```suggestion
    if (e.cause) {
      logger.fatal(e.cause.stack, "error");
    }
```

or can `cause.stack` be undefined?
```suggestion
    if (e.cause?.stack) {
      logger.fatal(e.cause.stack, "error");
    }
```

src/lib/fauna.mjs
ecooper: I can switch it. The problem is that this is library code and we should never really get this far without a query. You can't write a helpful validation message (i.e. use `[fql] or --input`) in this context.

ptpaterson:
> We do in query via --extra.
🤦Okay... I'm catching up.

src/lib/fauna.mjs
ptpaterson: trying a few things like leaving the query arg missing or providing an empty string catch the error before this point. So this may be moot? If the other code is working query should always be truthy by this point.

PR #463, opened 2024-12-04
group options

echo-bravo-yahoo:
right now, we present all the options in one blob. this change groups the config options into these groups:
- Output (format, quiet, color)
- Config (config, profile)
- API (user, role, database, secret)
- Debug (verbosity, verboseComponent)
- Options (help and all command-specific options)

these groups can/should be renamed and rearranged, but this grouping should be a readability improvement over the current soup of flags.

before:
image

after:
image

### DIFF
--- a/src/cli.mjs
+++ b/src/cli.mjs
@@ -120,11 +120,14 @@ function buildYargs(argvInput) {
         type: "boolean",
         // https://github.com/chalk/chalk?tab=readme-ov-file#chalklevel
         default: chalk.level > 0,
+        group: "Output:",
       },
       config: {
         type: "string",
-        description: "Path to a CLI config file to use. Use `--profile` to select a profile from the file.",
+        description:
+          "Path to a CLI config file to use. Use `--profile` to select a profile from the file.",
         default: ".",
+        group: "Config:",
       },
       profile: {
         alias: "p",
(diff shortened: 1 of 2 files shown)


src/cli.mjs
ptpaterson, line 1: What do you think about something like "Authentication/Authorization" instead of "API"? definitely more verbose... "AuthN/Z"?

src/cli.mjs
```diff
@@ -120,46 +120,58 @@ function buildYargs(argvInput) {
...
-        description: "Only output the results of the command. Useful for scripts, CI/CD, and automation workflows.",
+        description:
+          "Only output the results of the command. Useful for scripts, CI/CD, and automation workflows.",
         default: false,
+        group: "Output:",
       },
       verboseComponent: {
         description:
-          "Components to emit diagnostic logs for. Takes precedence over the `--verbosity` flag. Pass components as a comma-separate list, such as `--verboseComponent fetch, error`, or as separate flags, such as `--verboseComponent fetch --verboseComponent error`.",
+          "Components to emit diagnostic logs for. Takes precedence over the `--verbosity` flag. Pass components as a comma-separate list, such as `--verboseComponent fetch,error`, or as separate flags, such as `--verboseComponent fetch --verboseComponent error`.",
```
ptpaterson, line 163: I can't actually make `--verboseComponent fetch,error` work btw. A problem for later.
https://faunadb.atlassian.net/browse/FE-6176

ptpaterson (approved):
Thought on names, but otherwise LGTM

src/cli.mjs
echo-bravo-yahoo: turns out the short syntax is actually `--verboseComponent fetch error argv`. updated!

src/cli.mjs
echo-bravo-yahoo: here's what's in the group now:
image

i feel like database and url are only tenuously auth.

henryfauna (approved):

src/cli.mjs
```diff
@@ -158,46 +158,58 @@ function buildYargs(argvInput) {
...
-        description: "Only output the results of the command. Useful for scripts, CI/CD, and automation workflows.",
+        description:
+          "Only output the results of the command. Useful for scripts, CI/CD, and automation workflows.",
         default: false,
+        group: "Output:",
       },
       verboseComponent: {
         description:
-          "Components to emit diagnostic logs for. Takes precedence over the `--verbosity` flag. Pass components as a comma-separate list, such as `--verboseComponent fetch, error`, or as separate flags, such as `--verboseComponent fetch --verboseComponent error`.",
+          "Components to emit diagnostic logs for. Takes precedence over the `--verbosity` flag. Pass components as a space-separated list, such as `--verboseComponent fetch error`, or as separate flags, such as `--verboseComponent fetch --verboseComponent error`.",
```
jrodewig, line 201: Thanks for fixing this!

PR #464, opened 2024-12-04
[FE-6169] Support Shorthand Region Groups

henryfauna:
Ticket(s): FE-6169

## Problem
You must use the region groups that the account api expects as input to the `--database` arg. eg. `--database us-std/my-database`. We do not often expose the name `us-std` to customers and it will likely be more natural to use `--database us/my-database` as that is what is presented in the dashboard.

## Solution
Standardize region groups before they are passed to the account API. This allows the usage of `--database us-std/my-database` or `--database us/my-database`.

## Result
You can use whatever region group reference you prefer.

## Testing
Updated existing tests.

### DIFF
--- a/src/lib/fauna-account-client.mjs
+++ b/src/lib/fauna-account-client.mjs
@@ -220,17 +220,17 @@ export class FaunaAccountClient {
    * Transforms database paths to standardize region group naming conventions expected by
    * the account API.
    *
-   * @param {string} databasePath - The database path to standardize
-   * @returns {string} The standardized path
+   * @param {string} [databasePath] - The database path to standardize
+   * @returns {string | undefined} The standardized path
    * @example
    * // Returns "us-std/my-database"
-   * standardizeRegion("us/my-database")
+   * FaunaAccountClient.standardizeRegion("us/my-database")
    *
    * // Returns "eu-std/my-database"
-   * standardizeRegion("eu/my-database")
+   * FaunaAccountClient.standardizeRegion("eu/my-database")
    *
    * // Returns "global/my-database"
-   * standardizeRegion("classic/my-database")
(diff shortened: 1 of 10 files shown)


src/lib/fauna-account-client.mjs
```diff
@@ -204,12 +207,50 @@ export class FaunaAccountClient {
...
+   * // Returns "eu-std/my-database"
+   * standardizeRegion("eu/my-database")
+   *
+   * // Returns "global/my-database"
+   * standardizeRegion("classic/my-database")
+   *
+   * @throws {TypeError} If databasePath is provided but not a string
+   */
+  standardizeRegion(databasePath) {
+    if (!databasePath) return databasePath;
```
echo-bravo-yahoo, line 238: what case is this handling & should we prefer to throw instead?

src/lib/fauna-account-client.mjs
```diff
@@ -204,12 +207,50 @@ export class FaunaAccountClient {
...
+    const rest = parts.slice(1).join("/");
+
+    const regionMap = {
+      us: "us-std",
+      eu: "eu-std",
+      classic: "global",
+    };
+
+    const standardRegion = regionMap[region] || region;
+    return `${standardRegion}/${rest}`;
```
echo-bravo-yahoo, line 254: can you add a test (general-cli.mjs maybe?) that confirms the regions get mapped?

henryfauna: Added!

src/lib/fauna-account-client.mjs
henryfauna: I don't think so - an undefined or empty path is valid in some cases. For instance on the list database API.

mwilde345 (approved):

ecooper (approved):

PR #466, opened 2024-12-04
Add dynamic key support to schema

ecooper:
## Problem

Schema commands do not support dynamic database keys.

## Solution

Add support for dynamic database keys via `getSecret()` and `retryInvalidCredsOnce`.

## Result

```
➜  fauna-shell (v3-db-schema) ./src/user-entrypoint.mjs schema pull -d us-std/demo --dir /tmp/schema -p dev                                                                                                                                           ✱
Pull will make the following changes:
overwrite: collections.fsl
overwrite: functions.fsl
overwrite: main.fsl
✔ Accept the changes? yes
➜  fauna-shell (v3-db-schema) ls -la /tmp/schema/                                                                                                                                                                                                     ✱
total 32
drwxr-xr-x  5 ericcooper  wheel   160 Dec  4 12:45 .
drwxrwxrwt  5 root        wheel   160 Dec  4 12:45 ..
-rw-r--r--  1 ericcooper  wheel  2780 Dec  4 12:49 collections.fsl
-rw-r--r--  1 ericcooper  wheel  4539 Dec  4 12:49 functions.fsl
-rw-r--r--  1 ericcooper  wheel   263 Dec  4 12:49 main.fsl
➜  fauna-shell (v3-db-schema)      
```

## Testing

Existing tests should work as-is since this should be a transparent change.

## Follow-up
* Consider combining `getSecret` and `retryInvalidCredsOnce` into one wrapper.

### DIFF
--- a/src/lib/db.mjs
+++ b/src/lib/db.mjs
@@ -1,6 +1,8 @@
 //@ts-check
 
+
 import { container } from "../cli.mjs";
+import { retryInvalidCredsOnce } from "./fauna-client.mjs";
 
 function buildParamsString({ argv, params, path }) {
   const routesWithColor = ["/schema/1/staged/status", "/schema/1/validate"];
@@ -23,6 +25,7 @@ function buildParamsString({ argv, params, path }) {
  * @property {method} method - The HTTP method to use when making the request.
  * @property {object} [body] - The body to include in the request.
  * @property {boolean} [shouldThrow=true] - Whether or not to throw if the network request succeeds but is not a 2XX. If this is set to false, makeFaunaRequest will return the error instead of throwing.
+ * @property {string} [secret] - The secret to use when making the request.
  */
 
 /**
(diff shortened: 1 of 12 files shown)


henryfauna (approved):

src/lib/db.mjs
```diff
@@ -51,18 +57,34 @@ export async function makeFaunaRequest({
...
+
+/**
+ * @param {fetchParameters} opts
+ * @returns {Promise<Object>}
+ */
+export function makeRetryableFaunaRequest(opts) {
+  return retryInvalidCredsOnce(opts.secret, (secret) =>
+    makeFaunaRequest({ ...opts, secret }),
+  );
+}
```
echo-bravo-yahoo, line 90: not for this PR, but i think it'd be lower mental overhead to have just one entrypoint (`makeFaunaRequest`) and pass it `opts, argv`, and have it choose if it should be retryable or not. exporting multiple entrypoints and expecting consumers to pick correctly is more brittle.

echo-bravo-yahoo (approved):

PR #467, opened 2024-12-04
Consolidate Validation Helpers

henryfauna:
Ticket(s): FE-6162

## Problem
We have a `validateDatabaseOrSecret` helper and a `validateSecretOrDatabase` helper. We don't need both.

## Solution
Consolidate onto just one of these.

## Result
We have one place where we tell users they need to specify `--database`, `--secret` or `--local` for commands that need to talk to a specific database, eg `database create` or `database delete`.

### DIFF
--- a/src/lib/command-helpers.mjs
+++ b/src/lib/command-helpers.mjs
@@ -152,7 +152,7 @@ export function isUnknownError(error) {
 export const validateDatabaseOrSecret = (argv) => {
   if (!argv.database && !argv.secret && !argv.local) {
     throw new ValidationError(
-      "No database or secret specified. Please specify either --database, --secret, or --local to connect to your desired Fauna database.",
+      "No database or secret specified. Please use either --database, --secret, or --local to connect to your desired Fauna database.",
     );
   }
 };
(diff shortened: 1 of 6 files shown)


src/lib/command-helpers.mjs
```diff
@@ -151,7 +152,7 @@ export function isUnknownError(error) {
 export const validateDatabaseOrSecret = (argv) => {
   if (!argv.database && !argv.secret && !argv.local) {
     throw new ValidationError(
-      "No database or secret specified. Pass either --database, or --secret, or --local.",
+      "No database or secret specified. Please use either --database, --secret, or --local to connect to your desired Fauna database.",
```
echo-bravo-yahoo, line 155: would it make sense for this function to return true or throw? then you could, e.g., directly return it in validate because it returns true.

echo-bravo-yahoo (approved):

src/lib/command-helpers.mjs
henryfauna: yaa good idea

ecooper (approved):

PR #468, opened 2024-12-04
Create Key

cleve-fauna:
Ticket(s): FE-6128

## Problem

Creating a key is needed for the CLI. It is also needed to establish a --local pattern that works for many commands.

## Solution

In this PR do a basic create key implementation that uses the account API.

## Result
We got it started.
## Testing

Ran the full suite.


src/lib/fauna-account-client.mjs
```diff
@@ -198,20 +200,20 @@ export class FaunaAccountClient {
...
+  async createKey({ path, role, ttl, name }) {
+    return this.retryableAccountRequest({
       method: "POST",
       path: "/databases/keys",
       body: JSON.stringify({
         role,
         path: FaunaAccountClient.standardizeRegion(path),
-        ttl: ttl || new Date(Date.now() + TTL_DEFAULT_MS).toISOString(),
-        name: "System generated shell key",
+        ttl,
```
echo-bravo-yahoo, line 213: do we want to use the default TTL computed above (and commented out)? if so, let's; if not, let's trash the comment

src/lib/fauna-account-client.mjs
```diff
@@ -3,6 +3,8 @@
 import { container } from "../cli.mjs";
 import { InvalidCredsError } from "./misc.mjs";
 
+// const KEY_TTL_DEFAULT_MS = 1000 * 60 * 60 * 24;
+
```
cleve-fauna, line 7:
```suggestion
```

src/lib/misc.mjs
```diff
@@ -38,6 +38,23 @@ export function isTTY() {
...
 }
 
+/**
+ * Formats an object for display.
+ * @param {any} obj - The object to format
+ * @param {object} [opts] - Options
+ * @param {boolean} [opts.json] - Whether to return a JSON string with no pretty-printing
+ * @returns {string} The formatted object
+ */
+export function formatObject(obj, opts = {}) {
```
echo-bravo-yahoo, line 48: 👍 

are there few enough usages of formatObjectForShell that we can remove the export from it so this is the only easy-to-use formatting method?

src/commands/key/create.mjs
```diff
@@ -0,0 +1,77 @@
...
+        );
+      }
+      if (argv.database === undefined && argv.secret === undefined) {
+        throw new Error(
+          "You must provide at least one of: --database, --secret, --local.",
+        );
+      }
+      return true;
+    })
+    .help("help", "show help");
```
echo-bravo-yahoo, line 69:
```suggestion
```

echo-bravo-yahoo: this is handled at the top-level exclusively now - easier than remembering to provide help on every command :)

echo-bravo-yahoo (approved):

src/lib/misc.mjs
cleve-fauna: i don't want to touch any of this because there is a story to change all this for the formatting story.

src/lib/fauna-account-client.mjs
cleve-fauna: I think i deleted that already?

cleve-fauna: I don't see it anywyhere in this file: https://github.com/fauna/fauna-shell/blob/d84e9db87d9d014b3a8f21d77938b63525807698/src/lib/fauna-account-client.mjs

src/commands/key/create.mjs
```diff
@@ -0,0 +1,76 @@
...
+        description: "The role to assign to the key; e.g. admin",
+      },
+    })
+    .check((argv) => {
+      if (argv.ttl && !DateTime.fromISO(argv.ttl).isValid) {
+        throw new Error(
+          `Invalid ttl '${argv.ttl}'. Provide as an ISO 8601 date time string.`,
+        );
+      }
+      if (argv.database === undefined && argv.secret === undefined) {
```
henryfauna, line 62: I think we have a validation helper that does this already. I think it's called `validateDatabaseOrSecret`

henryfauna (approved):

src/commands/key/create.mjs
cleve-fauna: had to add this bc of tweaks i made in this command big dog.

cleve-fauna: if you odon't have this it asks you to login, which is insufficient.

PR #469, opened 2024-12-04
Align schema commands with public schema endpoints

jrodewig:
## Problem

We plan to publicly publish the Schema HTTP API endpoints. The CLI uses the endpoints in ways we don't want to support.

## Solution

Updates the CLI to align with our [planned docs](https://deploy-preview-2707--fauna-docs.netlify.app/fauna/current/reference/http/reference/core-api/#tag/Schema):

* Replace `/schema/1/validate` with its `/schema/1/diff` alias
* Change the `?diff` param to `?format`
* Change `?diff=true` to `?format=semantic`
* Remove usage of `?force=true`. This is the default if `version` isn't passed.
* Remove `text_diff` from `/schema/1/staged/status` responses.

## Testing

Updated the tests and ran `npm run test:local`. Happy to add additional tests if needed.

### DIFF
--- a/src/commands/schema/abandon.mjs
+++ b/src/commands/schema/abandon.mjs
@@ -9,20 +9,16 @@ async function doAbandon(argv) {
   const confirm = container.resolve("confirm");
 
   if (!argv.input) {
-    const params = new URLSearchParams({
-      force: "true", // Just abandon, don't pass a schema version through.
-    });
-
+    // Just abandon, don't pass a schema version through.
     await makeFaunaRequest({
       argv,
       path: "/schema/1/staged/abandon",
-      params,
       method: "POST",
     });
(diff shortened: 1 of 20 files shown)


echo-bravo-yahoo:
We'll definitely need to hand-test this change; right now the test suite doesn't have integration tests, and this is a change of the sort that really needs to stress integration w/the actual service.

jrodewig:
> We'll definitely need to hand-test this change; right now the test suite doesn't have integration tests, and this is a change of the sort that really needs to stress integration w/the actual service.

That's fair. To avoid delaying the RC, maybe we should shelf this and retry later with a separate bug bash. I'll write up a ticket and point back to this PR. Closing for now.

PR #472, opened 2024-12-04
Remove some options from login command

ptpaterson:
Ticket(s): [FE-6179](https://faunadb.atlassian.net/browse/FE-6179)

## Problem

`login` has all the options from other command, including `--user` which is described as setting the account that will be used to execute commands -- we want to use the `--user` option to set the name when registering the user. 

## Solution

Separate out "common" options from "common query" options and use those for the login command.

## Result

No changes other than to the help text. 

before
image

after
image

## Testing
no functional changes

[FE-6179]: https://faunadb.atlassian.net/browse/FE-6179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commands/login.mjs
+++ b/src/commands/login.mjs
@@ -39,7 +39,7 @@ function buildLoginCommand(yargs) {
     user: {
       alias: "u",
       type: "string",
-      description: "Name of account to register",
+      description: "User to log in as",
       default: "default",
       group: "login Options:",
     },
--- a/src/commands/login.mjs
+++ b/src/commands/login.mjs
@@ -1,7 +1,7 @@
 //@ts-check
 
 import { container } from "../cli.mjs";
-import { yargsWithCommonQueryOptions } from "../lib/command-helpers.mjs";
+import { yargsWithCommonOptions } from "../lib/command-helpers.mjs";
 import { FaunaAccountClient } from "../lib/fauna-account-client.mjs";
 
 async function doLogin() {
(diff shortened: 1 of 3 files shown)


echo-bravo-yahoo (approved):

src/lib/command-helpers.mjs
```diff
@@ -1,7 +1,37 @@
...
+  },
+};
+
 // used for queries customers can't configure that are made on their behalf
 const COMMON_QUERY_OPTIONS = {
+  user: {
+    alias: "u",
+    type: "string",
+    description:
+      "User account used to run the command. Register a user account in the CLI using `fauna login`.",
```
jrodewig, line 31:
```suggestion
      "User used to run the command. You must first log in as the user using `fauna login`.",
```

src/commands/login.mjs
```diff
@@ -35,7 +35,15 @@ async function doLogin() {
  * @returns
  */
 function buildLoginCommand(yargs) {
-  return yargsWithCommonQueryOptions(yargs);
+  return yargsWithCommonOptions(yargs, {
+    user: {
+      alias: "u",
+      type: "string",
+      description: "Name of account to register",
```
jrodewig, line 42:
```suggestion
      description: "User to log in as",
```

jrodewig (commented):
Left some help text suggestions. I've been iterating on these in the docs so this'll align them.

jrodewig (approved):

PR #473, opened 2024-12-04
Bare strings break colorize

ecooper:
## Problem

Bare strings would choke when colorizing.

0ba2e12a-9f6d-46de-a255-bd6b212c5737

## Solution

Stringify before colorizing.

Screenshot 2024-12-04 at 3 35 31 PM

### DIFF
--- a/src/lib/misc.mjs
+++ b/src/lib/misc.mjs
@@ -50,12 +50,12 @@ export function formatObjectForShell(obj, { color = true } = {}) {
     return JSON.stringify(obj, null, 2);
   }
 
-  return colorize(obj);
+  return colorize(JSON.stringify(obj));
 }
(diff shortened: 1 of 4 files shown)


echo-bravo-yahoo (approved):

src/lib/misc.mjs
```diff
@@ -67,5 +67,5 @@ export function formatFullErrorForShell(err, { color = true } = {}) {
     return JSON.stringify(err, null, 2);
   }
 
-  return colorize(err);
+  return colorize(JSON.stringify(err));
```
echo-bravo-yahoo, line 70: probably worth a quick regression test?

mwilde345 (approved):

PR #474, opened 2024-12-04
Don't let config parse for help or version when getting profile

ecooper:
## Problem

We parse args in the config middleware, but do not disable help or version. This means they will always execute before our actual help.

## Solution

Turn off help.

## Result

Help from the command show.

### DIFF
--- a/src/lib/config/config.mjs
+++ b/src/lib/config/config.mjs
@@ -80,13 +80,16 @@ export function configParser(path) {
 
   logger.debug(`Reading config from ${path}.`, "config");
   const config = getConfig(path);
-  const argv = yargs(argvInput).options({
-    profile: {
-      default: "default",
-      alias: ["p"],
-      type: "string",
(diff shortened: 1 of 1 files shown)


src/lib/config/config.mjs
```diff
@@ -80,13 +80,16 @@ export function configParser(path) {
...
+  const argv = yargs(argvInput)
+    .options({
+      profile: {
+        default: "default",
+        alias: ["p"],
+        type: "string",
+      },
+    })
+    .help(false)
+    .version(false).argv;
```
echo-bravo-yahoo, line 92: i think we should swap this out for yargs-parser instead

ecooper: That fails the test for some reason. I'll investigate.

henryfauna (approved):

echo-bravo-yahoo (approved):

PR #476, opened 2024-12-05
default oauth to prod client

mwilde345:
### Problem
CLI was defaulting to the oauth client that frontdoor uses when it's hosted locally. 
This means nobody should need to set any env vars when logging in to the shell, unless they wanna go to dev/prev

### Solution
Customers should default to prod. Remember this client secret isn't really secret and that's ok in this OAuth contract strategy.

Also dumping Fauna-prefixed env vars in the `logArgv` middleware for FE-6173

### Out of Scope

Option to redact specific env vars or argv attributes. Helpful for customers that want to copy pasta verbose output straight into a bug report.

### DIFF
--- a/src/lib/middleware.mjs
+++ b/src/lib/middleware.mjs
@@ -15,6 +15,16 @@ const DEFAULT_URL = "https://db.fauna.com";
 export function logArgv(argv) {
   const logger = container.resolve("logger");
   logger.debug(JSON.stringify(argv, null, 4), "argv", argv);
+  const faunaEnvVars = Object.entries(process.env)
+    .filter(([key]) => key.startsWith("FAUNA"))
+    .reduce((acc, [key, value]) => {
+      acc[key] = value;
+      return acc;
+    }, {});
+  logger.debug(
+    `Existing Fauna environment variables: ${JSON.stringify(faunaEnvVars)}`,
+    "argv",
+  );
   return argv;
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/middleware.mjs
```diff
@@ -15,6 +15,16 @@ const DEFAULT_URL = "https://db.fauna.com";
...
+  const faunaEnvVars = Object.entries(process.env)
+    .filter(([key]) => key.startsWith("FAUNA"))
+    .reduce((acc, [key, value]) => {
+      acc[key] = value;
+      return acc;
+    }, {});
+  logger.debug(
+    `Existing Fauna environment variables: ${JSON.stringify(faunaEnvVars)}`,
+    "argv",
+  );
```
mwilde345, line 27: example output

```
...
[argv]: Existing Fauna environment variables: {"FAUNA_ACCOUNT_URL":"https://account.fauna.com"}
[argv]: Defaulted url to 'https://db.fauna.com' no --url was provided
...
```

ptpaterson (approved):
this will help 👍

src/lib/middleware.mjs
cleve-fauna: I moved this to only caputre the env vars if user is in debug. otherwise we're revving through env vars on every command.

cleve-fauna (approved):

PR #481, opened 2024-12-05
FE-6181 validate profile when no config is found

ptpaterson:
Ticket(s): [FE-6181](https://faunadb.atlassian.net/browse/FE-6181)

## Problem

You can provide a profile when there is no config file and it will behave the same as `--profile default`. We should error instead.

## Solution

Remove the default value for `profile` option set in `cli.mjs`. The default of `"default"` is already set if there is a config file.

In the case where a path to a config file cannot be found, check if the `profile` option was specified by the caller and throw a validation error if so.

**other stuff:**

Removed the mutable global `argvInput`, which is only used by config parsing, and instead use a closure with it to create the config parser. Not strictly necessary by any means, but I have strong feelings about side effects like this, so please holler at me if the curried function appears more confusing than the exported variable.

docs

## Result

## Testing

[FE-6181]: https://faunadb.atlassian.net/browse/FE-6181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ecooper (approved):

src/lib/config/config.mjs
```diff
@@ -68,36 +90,68 @@ function validateConfig(profileName, profileBody, configPath) {
...
+  /**
+   * A parser to convert config files into appropriate command line arguments
+   *
+   * @param {string} path
+   * @returns {object}
+   */
+  function configParserWithArgs(path) {
+    let parsedPath = path;
+    let parsedProfile;
+    const logger = container.resolve("logger");
```
echo-bravo-yahoo, line 109: this looks more complicated than it needs to be - want to pair on it today? i think we can get rid of the inner function and shake about half the lines of code.

src/cli.mjs
```diff
@@ -172,7 +170,6 @@ function buildYargs(argvInput) {
         type: "string",
         description:
           "Profile from the CLI config file to use. Each profile specifies a set of CLI settings.",
```
jrodewig, line 172: Since there's no longer a separate default column value, but behavior remains the same:
```suggestion
          "Profile from the CLI config file to use. Each profile specifies a set of CLI settings. Defaults to the 'default' profile.",
```

src/lib/config/config.mjs
```diff
@@ -68,18 +90,45 @@ function validateConfig(profileName, profileBody, configPath) {
...
+    // specified or that the profile is "default"
+    const preConfigArgv = yargsParser(argvInput, {
+      alias: {
+        profile: ["p"],
+      },
+      string: ["profile"],
+    });
+    if (preConfigArgv.profile && preConfigArgv.profile !== "default") {
+      throw new ValidationError(
+        `Profile "${preConfigArgv.profile}" cannot be specified, because no config file found at "${path}". ` +
```
jrodewig, line 121:
```suggestion
        `Profile "${preConfigArgv.profile}" cannot be specified because there's no config file found at "${path}". ` +
```

test/config.mjs
```diff
@@ -360,7 +360,7 @@ describe("configuration file", function () {
           configToReturn: noDefaultConfig,
         });
       } catch (e) {}
-      const errorText = `No "default" profile found in config file at ${path.join(__dirname, "../dev.yaml")}. Either specify a profile with "--profile NAME" or add a "default" profile.`;
+      const errorText = `No "default" profile found in config file at "${path.join(__dirname, "../dev.yaml")}". Either specify a profile with "--profile NAME" or add a "default" profile.`;
```
jrodewig, line 363:
```suggestion
      const errorText = `No "default" profile found in the config file at "${path.join(__dirname, "../dev.yaml")}". Either specify a profile with "--profile NAME" or add a "default" profile.`;
```

test/config.mjs
```diff
@@ -385,6 +385,17 @@ describe("configuration file", function () {
       expect(stripAnsi(stderr.getWritten())).to.equal(message);
     });
 
+    it("exits with an error if a profile is specified and there is no config", async function () {
+      try {
+        await run(`eval "Database.all()" --profile nonexistent`, container);
+      } catch (e) {}
+
+      const errorText = `Profile "nonexistent" cannot be specified, because no config file found at "${path.join(__dirname, "..")}". Remove the profile, or provide a config file.`;
```
jrodewig, line 393:
```suggestion
      const errorText = `Profile "nonexistent" cannot be specified because there's no config file found at "${path.join(__dirname, "..")}". Remove the profile, or provide a config file.`;
```

jrodewig (commented):
Minor nits. I would like to include the default in the help text tho. Thanks!

jrodewig (approved):

PR #487, opened 2024-12-06
[FE-6183] Update Permissions on Credentials Files

henryfauna:
Ticket(s): FE-6183

## Problem
Credential files are readable by non-owners.

## Solution
Set them to `chmod 600` when the file is created.

## Result
Credential files are not readable by non-owners.

## Testing
Screenshot 2024-12-05 at 5 10 15 PM

### DIFF
--- a/src/lib/file-util.mjs
+++ b/src/lib/file-util.mjs
@@ -335,7 +335,7 @@ export function initHistoryStorage() {
   }
   const historyFile = path.join(historyDir, "history");
   if (!fileExists(historyFile)) {
-    fs.writeFileSync(historyFile, "");
+    fs.writeFileSync(historyFile, "", { mode: 0o600 });
   }
 
   return historyFile;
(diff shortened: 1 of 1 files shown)


ptpaterson (approved):
Credentials are clearly most important. Do we want to limit access for the shell history file as well? Genuine question -- I don't feel strongly that we do, but want to ask what others think.

No problem just merging this though.

echo-bravo-yahoo:
> Credentials are clearly most important. Do we want to limit access for the shell [history file](https://github.com/fauna/fauna-shell/blob/593d7546bf8f7265edec0b8f5ebf40a6fdf3375d/src/lib/file-util.mjs#L323-L355) as well? Genuine question -- I don't feel strongly that we do, but want to ask what others think.
> 
> No problem just merging this though.

I think we should (err on the side of security). Users may be passing confidential info as args to UDFs and such.

echo-bravo-yahoo (approved):

PR #490, opened 2024-12-06
Rename "extra" option to "raw"

ptpaterson:
## Problem

"raw" may not be a perfect name for this option, but I believe it is a step forward from "extra"

## Solution

rename the option and all the argv values from "extra" to "raw"

## Result

no functional changes, just the name.

## Testing

no functional changes, just the name.


ecooper (approved):
Maybe it's worth saying it's the data from the JS drivers?

I know there's been some confusion/expectation it would be the raw http body.

ptpaterson:
> Maybe it's worth saying it's the data from the JS drivers?
> I know there's been some confusion/expectation it would be the raw http body.

Like in the help description? Or via name convention?

I think it would be nice to have the option to return the wire protocol, but I otherwise like the consistency with returning the `data` field in the `QuerySuccess` object in whatever format you would have received without the --extra/--raw option.

PR #491, opened 2024-12-06
Command to start a local container

cleve-fauna:
Ticket(s): FE-5990

## Problem
Fauna's docker container gives users an easy way to implement a reliable inner loop development cycle. It makes it easy for them to develop schemas and programs against Fauna locally, without incurring any service costs.

Our support for the container in tooling is absent. User's have to find it and build scripts to fire it up.

## Solution
Take a first step to making the container more prevalent in the tooling by making a CLI command to start the container.

## Result
You can start a local Fauna with `fauna local`
## Testing
Ran the tests. Notice how I added a bunch for the container.

## Out of Scope
I need to add tests for different argv inputs, including incorrect inputs. I've put a generic skipped test to flag that for now.


package.json
```diff
@@ -79,7 +80,7 @@
     "pretest:ci": "npm run build:app",
     "test:ci": "mocha --forbid-only --recursive ./test --require ./test/mocha-root-hooks.mjs --reporter mocha-multi-reporters --reporter-options configFile=./test/config/reporter.json",
     "build": "npm run build:app && npm run build:sea",
-    "build:app": "esbuild --bundle ./src/user-entrypoint.mjs --platform=node --outfile=./dist/cli.cjs --format=cjs --inject:./sea/import-meta-url.js --define:import.meta.url=importMetaUrl --define:process.env.NODE_ENV=\\\"production\\\"",
+    "build:app": "esbuild --loader:.node=file --bundle ./src/user-entrypoint.mjs --platform=node --outfile=./dist/cli.cjs --format=cjs --inject:./sea/import-meta-url.js --define:import.meta.url=importMetaUrl --define:process.env.NODE_ENV=\\\"production\\\"",
```
ecooper, line 83: tl;dr: What does this do?

src/lib/docker-containers.mjs
```diff
@@ -0,0 +1,324 @@
...
+    await pullImage(imageName);
+  }
+  const logStream = await startContainer({
+    imageName,
+    containerName,
+    hostPort,
+    containerPort,
+  });
+  logger.stderr(
+    `[StartContainer] Container '${containerName}' started. Monitoring HealthCheck for readiness.`,
```
ecooper, line 31: FWIW, we could update `colorize` (and `codeToAnsi`) to support colorizing logs in a couple lines:

https://textmate-grammars-themes.netlify.app/?theme=github-dark-high-contrast&grammar=log

ecooper (approved):
Looking sharp.

I still want the ability to push a schema to a database (creating it if it doesn't already exist), but this is slick by itself too.

src/commands/local.mjs
```diff
@@ -0,0 +1,60 @@
...
+      describe:
+        "The port on the host machine mapped to the container's port. This is the port you'll connect to Fauna on.",
+      type: "number",
+      default: "8443",
+    },
+    image: {
+      describe: "The image to run locally",
+      type: "string",
+      default: "fauna/faunadb:latest",
+    },
```
ecooper, line 41: Why is this customizable?

src/commands/local.mjs
```diff
@@ -0,0 +1,60 @@
...
+      describe: "The name to give the container",
+      type: "string",
+      default: "faunadb",
+    },
+    pull: {
+      describe: "Pull the latest image before starting the container",
+      type: "boolean",
+      default: true,
+    },
+  });
```
ecooper, line 52: We can do this in a future PR, but I really love to see FSL directory flag and a database name flag that this command creates and pushes into.

package.json
cleve-fauna: Some node modules like low level crypto libraries esbuild does not bundle by default. This instructs esbuild to do so by copying the files it needs into the bundle.

src/lib/docker-containers.mjs
cleve-fauna: That's a good idea.

src/commands/local.mjs
cleve-fauna: It probably doesn't need to be. I put it in case someone wanted to use the "nightly" container build. But taht seems like an internal concern not appropriate for this tooling.

src/commands/local.mjs
cleve-fauna: That's good idea for starting up dev environments.

package.json
cleve-fauna: There are specific versions published: https://hub.docker.com/r/fauna/faunadb/tags

So i could see a case for having the tag be mutable for debugging. So you'd take a `--tag` argument instead. But we could also just drop it and always do latest which seems more appropriate for a web service.

src/commands/local.mjs
cleve-fauna: I removed this flag.

test/local.mjs
```diff
@@ -0,0 +1,212 @@
...
+    logsStub.callsFake(async () => ({
+      on: () => {},
+      destroy: () => {},
+    }));
+    docker.createContainer.resolves({
+      start: startStub,
+      logs: logsStub,
+      unpause: unpauseStub,
+    });
+    await run("local", container);
```
echo-bravo-yahoo, line 49: cool tests :)

src/commands/local.mjs
```diff
@@ -0,0 +1,55 @@
...
+      default: "8443",
+    },
+    hostPort: {
+      describe:
+        "The port on the host machine mapped to the container's port. This is the port you'll connect to Fauna on.",
+      type: "number",
+      default: "8443",
+    },
+    name: {
+      describe: "The name to give the container",
```
echo-bravo-yahoo, line 38:
```suggestion
      describe: "The name to give the container.",
```

src/commands/local.mjs
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Builds the yargs command for the local command
+ * @param {import('yargs').Argv} yargs The yargs instance
+ * @returns {import('yargs').Argv} The yargs instance
+ */
+function buildLocalCommand(yargs) {
+  return yargs.options({
+    containerPort: {
+      describe: "The port inside the container Fauna listens on",
```
echo-bravo-yahoo, line 27:
```suggestion
      describe: "The port inside the container Fauna listens on.",
```

i believe we're standardizing on periods always.

src/commands/local.mjs
```diff
@@ -0,0 +1,55 @@
...
+      type: "number",
+      default: "8443",
+    },
+    name: {
+      describe: "The name to give the container",
+      type: "string",
+      default: "faunadb",
+    },
+    pull: {
+      describe: "Pull the latest image before starting the container",
```
echo-bravo-yahoo, line 43:
```suggestion
      describe: "Pull the latest image before starting the container.",
```

src/lib/docker-containers.mjs
```diff
@@ -0,0 +1,325 @@
...
+
+/**
+ * Ensures the container is running
+ * @param {string} imageName The name of the image to create the container from
+ * @param {string} containerName The name of the container to start
+ * @param {number} hostPort The port on the host machine mapped to the container's port
+ * @param {number} containerPort The port inside the container Fauna listens on
+ * @param {boolean} pull Whether to pull the latest image
+ * @returns {Promise<void>}
+ */
```
echo-bravo-yahoo, line 14: love the JSDoc annotations, they're really polished / helpful

src/lib/docker-containers.mjs
```diff
@@ -0,0 +1,325 @@
...
+  logger.stderr(
+    `[StartContainer] Container '${containerName}' started. Monitoring HealthCheck for readiness.`,
+  );
+  await waitForHealthCheck({
+    url: `http://localhost:${hostPort}`,
+    logStream,
+  });
+  logger.stderr(
+    `[ContainerReady] Container '${containerName}' is up and healthy.`,
+  );
```
echo-bravo-yahoo, line 40: not for this PR, but this is probably a good fit for one of the verbosity level loggers (`logger.debug/info/warn/error/fatal`). maybe we put it on info and change the default log level to include info? going through the verbosity loggers gives users easier / better sifting (b/c you give it a component and a log level, and they can use those to filter in/out sets of logs).

PR #494, opened 2024-12-09
add debug logging to fauna client queries

mwilde345:
Ticket(s): FE-6221

## Problem
Couldn't repro what was tripping up our JSON parsing from this ticket: https://faunadb.atlassian.net/browse/FE-6221

Tried messing with short term creds

Would have liked to see the raw fauna query response before it got to our parsing, but the `verbosity` logging didn't help with that. 

## Solution

Added some debug logging to the raw query responses.

## Result

Can get better results from customer debug logs

## Testing

ran with `--verbosity 5` and saw the query response. But didn't get an error yet


src/lib/fauna.mjs
```diff
@@ -87,9 +87,24 @@ export const runQuery = async ({
...
+    .then((result) => {
+      logger.debug(
+        `Fauna v10 query result: ${JSON.stringify(result, null, 2)}`,
+        "client",
+      );
+      return result;
+    })
+    .catch((err) => {
+      logger.debug(
+        `Fauna v10 query error: ${JSON.stringify(err, null, 2)}`,
```
henryfauna, line 103: Is it possible that err would not be valid JSON?

echo-bravo-yahoo: Suspect so - fetch timeouts are often circular. We could try several strategies in a fallback w/try catch? Maybe:
- JSON.stringify
- util.inspect
- toString

mwilde345: the caught error should always be an object of some sort which stringify can handle. If we wanted JSON.parse here that would be more problematic

PR #495, opened 2024-12-09
[FE-6195] Disable JS Completions

henryfauna:
Ticket(s): FE-6195

## Problem
Setting `completer: undefined` when we set up the repl means we default to JS completions. This is why you can get things like:

Screenshot 2024-12-09 at 10 42 11 AM

`setTimeout` is of course not a function in Fauna. Previously the shell would parse the expression and catch the appropriate error code to avoid logging these sorts of errors. See here: https://github.com/fauna/fauna-shell/blob/1c5153c5d49076e0501a51c4b9ea0d2f5c2b8ecb/src/commands/shell.js#L109

## Solution
Disable JS completions as these don't make sense here.

## Testing
Behold typing `Time.` does not blow up:

Screenshot 2024-12-09 at 10 58 01 AM


mwilde345 (approved):

src/commands/shell.mjs
```diff
@@ -21,11 +21,11 @@ async function shellCommand(argv) {
 
   /** @type {import('node:repl').ReplOptions} */
   const replArgs = {
-    prompt: `${argv.db_path || ""}> `,
+    prompt: `${argv.database || ""}> `,
     ignoreUndefined: true,
     preview: argv.apiVersion !== "10",
     // TODO: integrate with fql-analyzer for completions
-    completer: argv.apiVersion === "10" ? () => [] : undefined,
+    completer: () => [],
```
echo-bravo-yahoo, line 28: So, we technically do lose functionality here vs the old CLI:
iTerm2 2024-12-09 12 00 19

I'm not sure I care, given that FQL v4 is on the deprecation path.

echo-bravo-yahoo (approved):

src/commands/shell.mjs
henryfauna: Yeah your right - let me take a closer look

echo-bravo-yahoo: the way the old code works is by making a nodeJS sub-vm with a context object that only contains the names defined in v4 syntax. so, setTimeout isn't a suggestion because setTimeout isn't defined as a global var in the sub-vm that evaluates completions.

https://nodejs.org/api/vm.html#scriptruninnewcontextcontextobject-options

henryfauna: Okay I think v4 completions should be back -

Screenshot 2024-12-09 at 4 40 42 PM

echo-bravo-yahoo: awesome, thank you for looking into it!

PR #496, opened 2024-12-09
Clean up output formatting and support colorizing FQL results

ecooper:
## Problem

We do not support copy-pasta FQL in shell and query.

## Solution

Use shiki for color support and add `codeToANSI` (aliased to `codeToAnsi` in the container so it's the same as strip). We are currently using `typescript` for highlights because the @shiki/cli only works with pre-bundled languages. A future PR will move the grammars over and replicate `codeToANSI`'s behavior. This is all surfaced through a `colorize` function. 

However by supporting FQL rendering, nearly all the tests broke. So while I was here:

* Added an `argv` dev command for testing when you just want to prove argv
* Delete the key commands since we were going to anyway so I didn't need to update their tests
* Update shell and query to support data formats for v10.

## Result

When outputting non-text, you can pass it to `colorize` with a format and it will format it.

## Testing

All tests needed to be updated to support fql by default.


src/commands/database/list.mjs
```diff
@@ -29,23 +29,24 @@ function pickOutputFields(databases, argv) {
...
-    container.resolve("logger").stdout(JSON.stringify(output));
-  } else {
-    container.resolve("logger").stdout(formatObjectForShell(output, { color }));
-  }
+  container.resolve("logger").stdout(
+    await colorize(output, {
+      format: JSON_FORMAT,
+      color: color,
+    }),
+  );
```
ptpaterson, line 45: This means that the json option is ignored for the list database list command, we'll just always return JSON, is that right?

src/commands/shell.mjs
```diff
@@ -21,11 +23,10 @@ async function shellCommand(argv) {
...
   /** @type {import('node:repl').ReplOptions} */
   const replArgs = {
-    prompt: `${argv.db_path || ""}> `,
+    prompt: `${argv.database ? argv.database : ""}> `,
     ignoreUndefined: true,
-    preview: argv.apiVersion !== "10",
-    // TODO: integrate with fql-analyzer for completions
-    completer: argv.apiVersion === "10" ? () => [] : undefined,
+    preview: undefined,
+    completer: undefined,
```
echo-bravo-yahoo, line 29: these can still support v4 if included/fixed, see one of @henryfauna's open PRs.

src/lib/fauna-client.mjs
```diff
@@ -150,9 +86,9 @@ export const runQueryFromString = (expression, argv) => {
  * @param {string} opts.apiVersion - The API version
  * @param {boolean} opts.extra - Whether to include extra information
  * @param {boolean} opts.color - Whether to colorize the error
- * @returns {object}
+ * @returns {Promise<string>}
  */
-export const formatError = (err, { apiVersion, extra, color }) => {
+export const formatError = async (err, { apiVersion, extra, color }) => {
```
echo-bravo-yahoo, line 91: why change to async?

ecooper: colorize is async because `codeToAnsi` is async.

shiki has a sync highlighter, but I'd need to reimplement the codeToAnsi stuff...which I'll do in a follow.

Then I can remove all the asyncz.

src/commands/database/list.mjs
ecooper: Yeah. There's not another format right now.

src/commands/shell.mjs
ecooper: Ah, I'll let his in first, and then resolve the conflicts so we don't lose his work.

mwilde345 (approved):

src/lib/fauna-client.mjs
ecooper: Async has been removed.

src/commands/shell.mjs
ecooper: This has been put back in.

henryfauna (approved):

cleve-fauna (commented):
Approve!

PR #499, opened 2024-12-10
Update use of schema API for the pull command

ptpaterson:
Ticket(s): [FE-6228](https://faunadb.atlassian.net/browse/FE-6228)

## Problem

We do not pass in the `version` or `staged` params to calls to get schema files.

Pulling active schema while a staged schema exists is possible, but we block it.

## Solution

Pass in the version and staged parameters to `/schema/1/files/{filename}` endpoint calls.

Update the confirmation message to indicate whether you are going to pull from active or staged schema.

## Result

We can pull schema, whether it's staged or not. We utilize OCC with schema versions correctly.

## Testing

Added tests that cover usage with the new behavior, including the --staged option.

[FE-6228]: https://faunadb.atlassian.net/browse/FE-6228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ecooper (approved):

src/commands/schema/pull.mjs
```diff
@@ -52,51 +53,80 @@ function logDiff({ argv, adds, overwrites, deletes }) {
...
     );
-  } else if (statusResponse.status === "none" && !argv.active) {
-    throw new CommandError("There are no staged schema changes to pull.");
   }
+  const source =
+    (argvSource ?? statusResponse.status === "none") ? "active" : "staged";
+
+  logger.debug(
+    `Pulling remote ${source} schema, version '${version}'.`,
+    "schema pull",
```
ecooper, line 122: @echo-bravo-yahoo do we have a preference of these being one word to simplify passing them?

mwilde345: yeah i don't think it would work as expected unless it was a single or hyphenated word

PR #500, opened 2024-12-10
handle accountKey arg with user and secret

mwilde345:
Ticket(s): FE-###

## Problem

## Solution

## Result

## Testing

### DIFF
--- a/src/lib/auth/credentials.mjs
+++ b/src/lib/auth/credentials.mjs
@@ -7,15 +7,24 @@ import { AccountKeys } from "./accountKeys.mjs";
 import { DatabaseKeys } from "./databaseKeys.mjs";
 
 const validateCredentialArgs = (argv) => {
-  if (argv.database && argv.secret && !argv.local) {
-    throw new ValidationError(
-      "Cannot use both the '--secret' and '--database' options together. Please specify only one.",
-    );
-  } else if (argv.role && argv.secret && !argv.local) {
-    // The '--role' option is not supported when using a secret. Secrets have an
-    // implicit role.
-    throw new ValidationError(
-      "The '--role' option is not supported when using a '--secret'. Please specify only one.",
+  const logger = container.resolve("logger");
+  const illegalArgCombos = [
+    ["accountKey", "secret", "local"],
+    ["secret", "database", "local"],
+    ["secret", "role", "local"],
+  ];
+  for (const [first, second, conditional] of illegalArgCombos) {
+    if (argv[first] && argv[second] && !argv[conditional]) {
+      throw new ValidationError(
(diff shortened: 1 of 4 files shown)


henryfauna (approved):

src/lib/auth/databaseKeys.mjs
```diff
@@ -50,7 +50,7 @@ export class DatabaseKeys {
     // argv.secret comes from flag, config, or FAUNA_SECRET
     if (argv.secret) {
       key = argv.secret;
-      keySource = "--secret";
+      keySource = "user";
```
ptpaterson, line 53: So this is this describing that the source of the key is a user-provided value? Not that specifically the `--user` arg was provided, correct?

src/lib/auth/credentials.mjs
```diff
@@ -7,15 +7,24 @@ import { AccountKeys } from "./accountKeys.mjs";
...
+    if (argv[first] && argv[second] && !argv[conditional]) {
+      throw new ValidationError(
+        `Cannot use both the '--${first}' and '--${second}' options together. Please specify only one.`,
+      );
+    }
+  }
+
+  if (argv.user && argv.accountKey) {
+    logger.debug(
+      "Both 'user' and 'account-key' arguments were specified. 'account-key' will be used to mint database secrets. 'user' will be ignored.",
```
jrodewig, line 26: We use `--accountKey` in the help, but it looks like we accept both. I'll update the docs, but it may be good if our help hints output both (or the hyphenated version if that's preferred).

src/lib/command-helpers.mjs
```diff
@@ -54,8 +54,10 @@ const COMMON_QUERY_OPTIONS = {
   },
   accountKey: {
     type: "string",
-    description: "The account key to use when calling Fauna",
+    description:
+      "The key to use for calling the Fauna Account API. Providing an account key will negate the need for a user login. The key will be used to generate short-lived database secrets. Cannot be used with --user or --secret.",
```
jrodewig, line 58: IMO we don't need to call out the Fauna Account API here. Some other wording nits to align with other help.
```suggestion
      "Fauna account key used for authentication. Negates the need for a user login. The key is used to generate short-lived database secrets for the CLI. Mutually exclusive with `--user` and `--secret`.",
```

jrodewig (approved):
Looks good. Thanks!

src/lib/auth/databaseKeys.mjs
mwilde345: correct

src/lib/auth/credentials.mjs
mwilde345: put it back to camelcase for simplicity

PR #501, opened 2024-12-10
Add performance hints

ecooper:
## Problem

The query API can enable performance hints in the payload summary. This feature is not available in the CLI.

## Solution

This PR adds performance hints to query and shell. Within shell, you can toggle them on and off. When using query, they are printed to stderr so they do not interfere with things reading outputs.

Because performance hints come through summary, some parsing logic needs to be in place to protect against that. Additionally, we apply fql formatting to the actual diagnostic text itself. If for whatever reason this fails, we just print the summary as is.

## Result

Query
Screenshot 2024-12-10 at 2 07 43 PM

Query w/ summary
Screenshot 2024-12-10 at 3 55 11 PM

Shell
Screenshot 2024-12-10 at 2 08 22 PM

Shell with multiple hints
Screenshot 2024-12-10 at 3 52 00 PM

## Testing

Tests were added to prove its wired up to the driver and displays appropriately in shell and query.


ptpaterson (approved):

src/commands/shell.mjs
```diff
@@ -105,11 +109,30 @@ async function shellCommand(argv) {
...
+        );
+        shell.prompt();
+      },
+    },
   ].forEach(({ cmd, ...cmdOptions }) => shell.defineCommand(cmd, cmdOptions));
 
   return completionPromise;
 }
 
+const getArgvOrCtx = (key, argv, ctx) => {
```
ptpaterson, line 128: I think we could get away with mutating argv for these shell commands to tweak settings, but I haven't gone too deep into what weird side effects that may create. I started looking into a `.setOption` command instead of a list of specific `.toggleX` commands, and was thinking that mutating argv might be necessary.

jrodewig (approved):

src/lib/fauna-client.mjs
```diff
@@ -122,3 +128,28 @@ export const formatQueryResponse = (
...
   }
 };
+
+/**
+ * Formats a performance hint. If no hint is available, returns a default message. If
+ * the hint is malformed, returns the hint as is.
+ * @param {string} performanceHint - The performance hint
+ * @returns {string}
+ */
+export const formatPerformanceHint = (performanceHint) => {
```
fauna-chase, line 138: we currently just pass this the entire summary right? the summary can contain multiple performance hints

ecooper: This has been updated to support showing the summary in general as well as rendering multiple performance hints.

Screenshot 2024-12-10 at 3 52 00 PM

fauna-chase (approved):

PR #502, opened 2024-12-10
Update usage of the schema API for push, status, and diff

ptpaterson:
Ticket(s): [FE-6236](https://faunadb.atlassian.net/browse/FE-6236)

## Problem

We should use the /diff endpoint rather than /validate. Also, /diff does not accept a force param.

## Solution

update requests to use the /diff endpoint. Remove the `force` param from `diff`.

## Result

No user-facing changes.

## Testing

Updated expectations in tests.

[FE-6236]: https://faunadb.atlassian.net/browse/FE-6236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henryfauna (approved):

jrodewig (approved):
🎉  Thanks for updating this.

PR #505, opened 2024-12-11
Handle No Local Schema Files

henryfauna:
Ticket(s): FE-6217

## Problem
When no local schema files are found `schema status` tells the user that everything will be removed.

## Solution
Detect if no files were found and output a message to that effect instead.

## Result
Users will know that they have specified the wrong schema directory.

Screenshot 2024-12-10 at 4 57 12 PM

## Testing
Updated & added tests.

### DIFF
--- a/src/commands/schema/status.mjs
+++ b/src/commands/schema/status.mjs
@@ -19,11 +19,11 @@ async function doStatus(argv) {
   const secret = await getSecret();
   const absoluteDirPath = path.resolve(argv.dir);
   const gatherFSL = container.resolve("gatherFSL");
-  const fsl = reformatFSL(await gatherFSL(argv.dir));
+  const fslFiles = await gatherFSL(argv.dir);
+  const hasLocalSchema = fslFiles.length > 0;
+  const fsl = reformatFSL(fslFiles);
 
-  const hasLocalSchema = fsl.entries().next().done === false;
-
-  const statusParams = new URLSearchParams({ diff: "summary" });
+  const statusParams = new URLSearchParams({ format: "summary" });
   const statusResponse = await makeFaunaRequest({
     argv,
     path: "/schema/1/staged/status",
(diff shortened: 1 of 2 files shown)


src/commands/schema/status.mjs
```diff
@@ -15,33 +16,38 @@ async function doStatus(argv) {
   const logger = container.resolve("logger");
   const makeFaunaRequest = container.resolve("makeFaunaRequest");
 
-  let params = new URLSearchParams({ diff: "summary" });
   const secret = await getSecret();
+  const absoluteDirPath = path.resolve(argv.dir);
   const gatherFSL = container.resolve("gatherFSL");
   const fsl = reformatFSL(await gatherFSL(argv.dir));
 
+  const hasLocalSchema = fsl.entries().next().done === false;
```
ptpaterson, line 24: It might be simpler to check the results of `gatherFSL` instead. Does the same thing, though.

```js
  const gatherFSL = container.resolve("gatherFSL");
  const fslFiles = await gatherFSL(argv.dir)
  const fsl = reformatFSL(fslFiles);

  const hasLocalSchema = fslFiles.length > 0;
```

src/commands/schema/status.mjs
```diff
@@ -15,33 +16,38 @@ async function doStatus(argv) {
...
+
+  const statusParams = new URLSearchParams({ diff: "summary" });
```
ptpaterson, line 26: There is no `diff` param documented for `/status` or `/diff` endpoints. Is the `format: "summary"` the param you want?

jrodewig: Can confirm — `format` is the new alias for `diff`.

src/commands/schema/status.mjs
henryfauna: nice

ecooper (approved):

test/schema/status.mjs
```diff
@@ -66,9 +75,52 @@ describe("schema status", function () {
...
+
+  it("notifies the user when no local schema is found", async function () {
+    gatherFSL.resolves([]);
+    fetch.onCall(0).resolves(
+      f({
+        version: 0,
+        status: "none",
+        diff: "Staged schema: none",
+        pending_summary: "",
+        text_diff: "",
```
jrodewig, line 89: The `/staged/status` endpoint doesn't return `text_diff` anymore.
```suggestion
```

jrodewig (approved):
Looks good aside from a minor test suggestion.

PR #506, opened 2024-12-11
Render errors as-is if they already have ansi characters

ecooper:
## Problem

If a command throws an error message that has already been formatted for display, we clobber it and make it red.  This means some errors, such as query syntax errors, lose the benefit of code highlighting.

## Solution

Detect if a thrown error has been formatted for display already, and don't apply `chalk.red` if it has.

While I was here, I moved the error handling code into a single module for discoverability and eventually testing.

## Result

Doesn't have ansi so it renders red:
Screenshot 2024-12-11 at 9 09 47 AM

Has ansi so we display as-is:
Screenshot 2024-12-11 at 9 10 19 AM


mwilde345 (approved):

ptpaterson (approved):
awesome.

PR #507, opened 2024-12-11
Notify on `schema push` if no schema files are found

henryfauna:
Ticket(s): FE-6218

## Problem
`schema push` will let you nuke your schema if you accidentally specify the wrong directory with no .fsl files.

## Solution
In order to run `schema push` you must specify a directory with .fsl files.

## Result
You can't run `fauna schema push --dir /wrong/dir --no-input` and nuke your schema.

## Testing
Added a test.


ecooper (approved):
Do we think this is a validationerror? Should we throw that?

The schema commands are interactive by design so I'm torn.

henryfauna:
> Do we think this is a validationerror? Should we throw that?
> 
> The schema commands are interactive by design so I'm torn.

Yeah I think an error makes sense here. It would be nice to not fail silently if you were running `schema push --no-input` from a script or something.

mwilde345 (approved):

PR #509, opened 2024-12-11
Fail the shell fast if you cannot query the database

ecooper:
## Problem

If you have an invalid database or something wacky with your secret, you won't know until you run the first command in your shell.

## Solution

Instead, do a quick queryable check and throw a validation error if the db is unqueryable for whatever reason. This will also exercise creds checking so that will fail fast too.

Additionally, 404s from frontdoor are thrown as `ValidationError` now since that happens outside the scope of `isQueryable`.

## Result

Shell will fail immediately if you cannot query the database or with the secret you've provided.

## Testing

`isQueryable` is injected into the container, and we have a unit test to prove things fail.


ptpaterson (approved):
Yes, please, and thank you!

`isQueryable` sounds like it ought to return a boolean, but that's a pretty pedantic nit.

src/lib/fauna-client.mjs
```diff
@@ -87,6 +88,20 @@ export const runQueryFromString = (expression, argv) => {
...
 
+/**
+ * Check if a database can be queried based on the current arguments.
+ * If it can't, it will throw an error.
+ * @param {*} argv
+ */
+export const isQueryable = async (argv) => {
+  const runQueryFromString = container.resolve("runQueryFromString");
+  try {
+    await runQueryFromString("1+1", argv);
```
cleve-fauna, line 99: big ❤️

cleve-fauna (approved):

PR #511, opened 2024-12-12
Redact secrets in logArgv

ecooper:
## Problem

When we log argv via verbosity, we log secrets. This is probably fine, but we also want to avoid people accidentally posting secrets to github issues where they use verbosity to debug.

## Solution

Add a custom stringify function that redacts the value of any key with secret or accountKey in it.

I think this is the only place we output secrets or keys.

## Result
Screenshot 2024-12-11 at 4 40 21 PM

## Testing

Added unit tests for `redact` and `redactedStringify`


echo-bravo-yahoo (approved):

henryfauna (approved):

src/lib/formatting/redact.mjs
```diff
@@ -0,0 +1,50 @@
...
+export function redact(text) {
+  if (!text) return text;
+
+  if (text.length < 12) {
+    return "*".repeat(text.length);
+  }
+
+  const lastFour = text.slice(-4);
+  const redactedLength = text.length - 4;
+  return "*".repeat(redactedLength) + lastFour;
```
cleve-fauna, line 17: might be good to put some of the prefix in too as they can hint to us what type of key it is.

See: https://github.com/fauna/backup-restore-frontdoor/blob/main/src/auth/constants.ts

For some examples

cleve-fauna: Honestly, i'd rather just do the prefix and redact all the rest. that way none of the random data is displayed. its just the envelope flag.

src/lib/formatting/redact.mjs
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param {string} text - The string to redact.
+ * @returns {string} The redacted string.
+ */
+export function redact(text) {
+  if (!text) return text;
+
+  if (text.length < 12) {
+    return "*".repeat(text.length);
+  }
```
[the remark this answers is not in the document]
ecooper: Was based on https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls

Figured it's effectively equivalent for this exercise.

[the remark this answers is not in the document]
cleve-fauna: Perfect. What i was looking for.

src/lib/formatting/redact.mjs
ecooper: That won't help a user figure out what secret was used if it's not something they expect (such as an expected profile or env passing secret). How about for strings larger than 16, we keep the first and last 4 and redact the middle?

PR #514, opened 2024-12-12
don't use a default profile

mwilde345:
Ticket(s): FE-6232

## Problem
If we find a config file in the default location, we assume they want to use it with at least the "default" profile, but they may not always want to use the config.

## Solution
Provide a way to not use the config, by requiring a profile if a config is found. Essentially removing the "default" profile

## Result
1. Config is provided (via flag or env) -> require profile via flag or env.
2. Config is not provided but is present in default location -> ignore the file if no profile is specified.
3. Config is not provided and not in default location -> error if profile is provided

## Testing
Updated existing tests

added test to ensure default config is ignored properly


src/lib/config/config.mjs
```diff
@@ -99,49 +99,65 @@ function validateConfig(profileName, profileBody, configPath) {
...
+
+  const profile = argv.profile || process.env.FAUNA_PROFILE;
+
+  // console.log(
+  //   "in config",
+  //   argvInput,
+  //   userProvidedConfigPath,
+  //   profile,
+  //   parsedPath,
+  // );
```
ptpaterson, line 126: oops! 😄

cleve-fauna (approved):

ptpaterson (approved):

PR #516, opened 2024-12-12
Clarify `--raw` flag help

jrodewig:
Ticket(s): [FE-6234](https://faunadb.atlassian.net/browse/FE-6234)

## Problem

The current `--raw` and `.toggleRawResponses` help text mentions the API response. This is misleading as the response actually comes from the JS driver.

## Solution

Rewords the text to avoid mentioning "full API response". 

## Result

Meh...? We could mention the JS driver here, but I think that gets very implementation-detail-y. This is mostly helpful for folks looking for query stats.

## Testing

N/A

[FE-6234]: https://faunadb.atlassian.net/browse/FE-6234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commands/query.mjs
+++ b/src/commands/query.mjs
@@ -166,7 +166,7 @@ function buildQueryCommand(yargs) {
       raw: {
         type: "boolean",
         description:
-          "Output the full API response, including summary and query stats.",
+          "Output the raw JSON query response, including summary and query stats.",
         default: false,
       },
     })
--- a/src/commands/shell.mjs
+++ b/src/commands/shell.mjs
@@ -104,7 +104,7 @@ async function shellCommand(argv) {
     },
     {
       cmd: "toggleRawResponses",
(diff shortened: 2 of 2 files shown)


ptpaterson:
The "simple" format is what is returned from the API and displayed. I don't think it is inaccurate to say API response body.

jrodewig:
> The "simple" format is what is returned from the API and displayed without additional processing. I don't think it is inaccurate to say API response body.

I don't think it's related to data format. Based on https://faunadb.atlassian.net/browse/FE-6234, we omit stuff like `error.message` (because it's omitted in the JS driver).

ptpaterson (approved):
Okay thank you for clarifying. I'm good with this. I am still trying to work out the final behavior of --raw (or whatever it may evolve into).

PR #519, opened 2024-12-12
Handle network errors with better messaging

ecooper:
## Problem

When a network error occurs in the v10 driver, we get a "The network encountered a problem" message. If fetch fails, you get a "fetch failed". Neither of these are helpful.

## Solution

Catch them with slightly better messaging and centralize the messaging so we are consistent in the places we need to handle these errors. Given the most likely cause of a network error is `--local`, add a blurb about that too.

All commands are covered by this messaging, but may display them differently depending on their error handling.

## Result

Screenshot 2024-12-12 at 11 17 57 AM

## Testing

Unit tests.

### DIFF
--- a/src/lib/errors.mjs
+++ b/src/lib/errors.mjs
@@ -6,6 +6,11 @@ import { container } from "../cli.mjs";
 
 const BUG_REPORT_MESSAGE = `If you believe this is a bug, please report this issue on GitHub: https://github.com/fauna/fauna-shell/issues`;
 
+// This error message is used in a few places where we handle network errors.
+export const NETWORK_ERROR_MESSAGE =
+  "Unable to connect to Fauna due to a network error. If you're using --local, " +
+  "make sure your local docker container is currently running with this command: fauna local";
+
 /*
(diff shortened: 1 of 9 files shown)


src/lib/errors.mjs
```diff
@@ -6,6 +6,11 @@ import { container } from "../cli.mjs";
 
 const BUG_REPORT_MESSAGE = `If you believe this is a bug, please report this issue on GitHub: https://github.com/fauna/fauna-shell/issues`;
 
+// This error message is used in a few places where we handle network errors.
+export const NETWORK_ERROR_MESSAGE =
+  "Unable to connect to Fauna due to a network error. If you're using --local, " +
+  "make sure your local docker container is currently running with this command: fauna local";
```
jrodewig, line 12:
```suggestion
  "Unable to connect to Fauna due to a network error. If using --local, " +
  "ensure your container is running with this command: fauna local";
```

jrodewig (approved):
LGTM. Left some message nits you can take or leave as you see fit.

PR #524, opened 2024-12-13
Color log lines in fauna local command.

cleve-fauna:
## Problem

We don't color log lines. This can make them harder to read and look less appealing.
## Solution

Add a highlighter for log lines. Apply it to the local command.

## Result

The local command colors logs. Other commands can add such coloring as needed.

## Testing
I ran the command and saw color:
Screenshot 2024-12-13 at 11 13 27 AM


src/lib/docker-containers.mjs
```diff
@@ -392,10 +388,19 @@ async function waitForHealthCheck({
...
   throw new CommandError(
     `[HealthCheck] Fauna at ${url} is not ready after ${maxAttempts} attempts. Consider increasing --interval or --maxAttempts.`,
   );
 }
+
+/**
+ * Outputs to stderr.
+ * @param {string} log The log
+ */
+function stderr(log) {
```
ptpaterson, line 403: Can we avoid the mutable global `color` by accepting it as an argument in `stderr`?

cleve-fauna: then i would have to pass color to every private method in this module.

I'd prefer to memoize it.

ptpaterson: Understood. I realize I prefer verbosity over pit traps like global variables. You could curry the color `const stderrColored = stderr.bind(null, color)` if you want, but you probably don't 😜

ptpaterson (approved):

src/lib/docker-containers.mjs
ptpaterson: no that's dumb in this case ignore that.

But you could declare stderr function within `ensureContainerRunning` to close over the color provided.

```js
export async function ensureContainerRunning({
  containerName,
  hostIp,
  hostPort,
  containerPort,
  pull,
  interval,
  maxAttempts,
  color,
}) {
  function stderr(log) {  
    logger.stderr(colorize(log, { format: Format.LOG, color }));
  }

  // ...
```

cleve-fauna: then i'd have to pass that around.

ptpaterson: misread the diff, it's needed in several functions 👍🏻

mwilde345 (approved):

PR #525, opened 2024-12-13
Optionally, create a db when starting a local container

cleve-fauna:
## Problem

Inner dev loops will be fastest if the user can:

1. start a Fauna container
2. create a database in that container
3. deploy a schema to that database

with a single command.

Right now we only support (1).

## Solution

Implement support for (2).

Do so by creating the requested DB if it doesn't exist, succeeding if it already exists with desired properties, failing if it exists but with different properties.
## Result

We got some tooling in place for supporting inner dev loops.
## Testing

Ran the tests; including new ones for happy and failure paths for creating a db.

### DIFF
--- a/src/commands/local.mjs
+++ b/src/commands/local.mjs
@@ -1,5 +1,10 @@
+import chalk from "chalk";
+import { AbortError } from "fauna";
+
+import { container } from "../cli.mjs";
 import { ensureContainerRunning } from "../lib/docker-containers.mjs";
 import { CommandError } from "../lib/errors.mjs";
+import { colorize, Format } from "../lib/formatting/colorize.mjs";
 
 /**
  * Starts the local Fauna container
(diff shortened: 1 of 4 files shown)


src/commands/local.mjs
```diff
@@ -80,6 +165,24 @@ function buildLocalCommand(yargs) {
...
+        );
+      }
+      if (argv.protected && !argv.database) {
+        throw new CommandError(
+          "--protected can only be set if --database is set.",
+          { hideHelp: false },
+        );
+      }
+      if (argv.priority && !argv.database) {
+        throw new CommandError(
```
echo-bravo-yahoo, line 181: Thoughts on making these `ValidationError`?

echo-bravo-yahoo (approved):

src/commands/local.mjs
cleve-fauna: Sure.

ecooper (approved):

PR #526, opened 2024-12-13
Print included query info fields as yaml

ptpaterson:
Ticket(s): [FE-6247](https://faunadb.atlassian.net/browse/FE-6247)

## Problem

As a user, we want to be able to view various query info fields when running queries.

## Solution

This PR add a new common query option called `include`, which can be an array of one or more of the following:
```
["all", "txnTs", "schemaVersion", "summary", "queryTags", "stats"]
```

If "all" is specified, all possible info fields should be displayed.

If `--performanceHints` is used, then at least the summary should be displayed. Don't worry if there are other components to the summary.

The query info fields should displayed so that they can be read by people. Most importantly, the summary field should be displayed plainly, formatted the way it comes back from the db. That is, don't show a weird string full of escape characters.

I used the `yaml` package to format the query info as an object and loaded the yaml colorizer. If color is on, we also apply FQL colorizing to appropriate summary lines.

## Result

query
image

shell
image

shell with toggling
image

image

## Testing

All existing tests pass.

WIP: will add some more tests

[FE-6247]: https://faunadb.atlassian.net/browse/FE-6247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/command-helpers.mjs
+++ b/src/lib/command-helpers.mjs
@@ -156,13 +156,7 @@ export function yargsWithCommonConfigurableQueryOptions(yargs) {
     }
 
     if (argv.include.includes("all")) {
-      argv.include = [
-        "txnTs",
-        "schemaVersion",
-        "summary",
-        "queryTags",
-        "stats",
-      ];
+      argv.include = [...QUERY_INFO_CHOICES];
     }
 
     if (argv.performanceHints && !argv.include.includes("summary")) {
--- a/src/lib/command-helpers.mjs
+++ b/src/lib/command-helpers.mjs
@@ -125,7 +125,7 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
   performanceHints: {
     type: "boolean",
     description:
-      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets the '--includes summary'",
+      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets '--include summary'",
     default: false,
     group: "API:",
   },
@@ -133,7 +133,7 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
     type: "array",
     choices: ["all", ...QUERY_INFO_CHOICES],
     default: [],
-    describe: "Select additional query information to include in the output",
+    describe: "Include additional query response data in the output.",
   },
 };
 
--- a/src/lib/command-helpers.mjs
+++ b/src/lib/command-helpers.mjs
@@ -114,19 +114,18 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
     default: 5000,
     group: "API:",
   },
-  summary: {
-    type: "boolean",
-    description:
-      "Output the summary field of the API response or nothing when it's empty. Only applies to v10 queries.",
-    default: false,
-    group: "API:",
-  },
   performanceHints: {
     type: "boolean",
     description:
-      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries.",
+      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets the '--includes summary'",
     default: false,
     group: "API:",
+  },
+  include: {
+    type: "array",
+    choices: ["all", "txnTs", "schemaVersion", "summary", "queryTags", "stats"],
+    default: [],
+    describe: "Select additional query information to include in the output",
   },
 };
 
@@ -135,7 +134,24 @@ export function yargsWithCommonQueryOptions(yargs) {
 }
 
 export function yargsWithCommonConfigurableQueryOptions(yargs) {
-  return yargsWithCommonOptions(yargs, COMMON_CONFIGURABLE_QUERY_OPTIONS);
+  return yargsWithCommonOptions(
+    yargs,
+    COMMON_CONFIGURABLE_QUERY_OPTIONS,
+  ).middleware((argv) => {
+    if (argv.include.includes("all")) {
+      argv.include = [
+        "txnTs",
+        "schemaVersion",
+        "summary",
+        "queryTags",
+        "stats",
+      ];
+    }
+
+    if (argv.performanceHints && !argv.include.includes("summary")) {
+      argv.include.push("summary");
+    }
+  });
 }
 
 export function yargsWithCommonOptions(yargs, options) {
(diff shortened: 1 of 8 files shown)


src/lib/command-helpers.mjs
```diff
@@ -114,28 +114,44 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
...
-      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries.",
+      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets the '--includes summary'",
     default: false,
     group: "API:",
   },
+  include: {
+    type: "array",
+    choices: ["all", "txnTs", "schemaVersion", "summary", "queryTags", "stats"],
+    default: [],
+    describe: "Select additional query information to include in the output",
```
jrodewig, line 128: You can take or leave the suggestion. I'd add a trailing period tho.
```suggestion
    describe: "Include additional query response data in the output.",
```

src/lib/command-helpers.mjs
```diff
@@ -114,28 +114,44 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
...
-    description:
-      "Output the summary field of the API response or nothing when it's empty. Only applies to v10 queries.",
-    default: false,
-    group: "API:",
-  },
   performanceHints: {
     type: "boolean",
     description:
-      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries.",
+      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets the '--includes summary'",
```
jrodewig, line 120:
```suggestion
      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets '--include summary'",
```

jrodewig (approved):

src/lib/command-helpers.mjs
```diff
@@ -114,28 +122,44 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
...
-    group: "API:",
-  },
   performanceHints: {
     type: "boolean",
     description:
-      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries.",
+      "Output the performance hints for the current query or nothing when no hints are available. Only applies to v10 queries. Sets '--include summary'",
     default: false,
     group: "API:",
   },
```
echo-bravo-yahoo, line 131: don't love this flag & behavior. any reason to not rely solely on `--include summary`?

src/lib/command-helpers.mjs
```diff
@@ -114,28 +122,44 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
...
+  include: {
+    type: "array",
+    choices: ["all", ...QUERY_INFO_CHOICES],
```
echo-bravo-yahoo, line 134: 👍

src/lib/command-helpers.mjs
```diff
@@ -114,28 +122,44 @@ const COMMON_CONFIGURABLE_QUERY_OPTIONS = {
...
+    default: [],
```
echo-bravo-yahoo, line 135: this might be spicy, but is this major version upgrade the right time to change the default behavior?

if we're targeting primarily human users, should it be `[summary, stats]`?

src/lib/fauna-client.mjs
```diff
@@ -179,3 +179,70 @@ export const formatQuerySummary = (summary) => {
...
+    if (Object.keys(queryInfoToDisplay).length === 0) return "";
+
+    const SUMMARY_IN_QUERY_INFO_FQL_REGEX = /^(\s\s\s\s\|)|(\d\s\|)/;
+    const colorized = colorize(queryInfoToDisplay, {
+      color,
+      format: Format.YAML,
+    })
+      .split("\n")
+      .map((line) => {
+        if (!line.match(SUMMARY_IN_QUERY_INFO_FQL_REGEX)) {
```
echo-bravo-yahoo, line 237: how about using the summary property & has-ansi to decide if we should colorize each block of queryInfoToDisplay? hopefully it'll be less brittle to summary additions/changes and any changes we make to our rendering of it.

e.g.:
```
let infoString = ""
for (const key of Object.keys(queryInfoToDisplay)) {
  key === 'summary'
    ? infoString += colorize(queryInfoToDisplay, { color, format: Format.FQL })
    : infoString += colorize(queryInfoToDisplay, { color, format: Format.YAML })
}
```

src/lib/command-helpers.mjs
ecooper: To enable performance hints in the summary, you need to set a special header on the request. This flag does that.

src/lib/command-helpers.mjs
ptpaterson: hmm I think I could get behind summary as a default. most queries don't have a summary, but having it on would make sure things like `dbg`, `log`, and any warning spit out by core appear by default.

I definitely want to se stats some times. I definitely don't want to see every time. Maybe

Shall we run with default as `["summary"]`?

ptpaterson: I will be switching over to doing more in CLI for random sandbox/test queries where I would most often go to the dashboard. That's the one case where I might want stats more often than not. I'll probably configure that profile in my global config to show stats every time.

src/lib/fauna-client.mjs
ptpaterson: It's a good call out on how brittle this bit is. 🤔 

We could split each key out more like
```js
key === 'summary'
  ? infoString += colorize({ [key]: queryInfoToDisplay[key], { color, format: Format.YAML })
  : infoString += formatQuerySummaryWithHeaderBits(queryInfoToDisplay.summary)
```

But check out the existing `formatQuerySummary` function. The first lines should NOT be formatted as FQL. Only the FQL hints should be, making formatting the summary quite an odd ball.

Out of scope for this PR is piping the included fields into ServiceError handling. Currently, rendering summary for errors is different than for successful queries. I think that's okay as is, but I don't know what it should be like for errors if we start rendering more info like stats.

I'd like to circle back to this once this is out and once Eric's error handling updates are in too. There may be some more common ground for rendering summary in a more reusable way. Even if not, following up to make sure summary rendering doesn't accidentally clobber rendering other output is smart. I'll file something to make sure we don't miss it, but I don't want to overdo it now while other things are in flux. Fair?

src/lib/command-helpers.mjs
ptpaterson: with default and "none" option
image

ecooper (approved):

PR #528, opened 2024-12-13
Fauna local no daemon text

echo-bravo-yahoo:
Ticket(s): FE-6238

before:
image

after:
image

### DIFF
--- a/src/lib/docker-containers.mjs
+++ b/src/lib/docker-containers.mjs
@@ -28,16 +28,28 @@ export async function ensureContainerRunning({
   color: _color,
 }) {
   color = _color;
-  if (pull) {
-    await pullImage(IMAGE_NAME);
+  let logStream;
+  try {
+    if (pull) {
+      await pullImage(IMAGE_NAME);
+    }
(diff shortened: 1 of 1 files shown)


cleve-fauna (changes requested):
Instead of this, which might not be portable, how about we use the ping API:

https://docs.docker.com/reference/api/engine/version/v1.37/#tag/System/operation/SystemPing

dockerode exposes this via `docker.ping()` which is an async function

https://github.com/apocas/dockerode/tree/master?tab=readme-ov-file#docker

echo-bravo-yahoo:
> Instead of this, which might not be portable, how about we use the ping API:
> 
> https://docs.docker.com/reference/api/engine/version/v1.37/#tag/System/operation/SystemPing
> 
> dockerode exposes this via `docker.ping()` which is an async function
> 
> https://github.com/apocas/dockerode/tree/master?tab=readme-ov-file#docker

gotcha - sounds good to me!

echo-bravo-yahoo:
relevant change is already in main - closing.

PR #529, opened 2024-12-13
Clean up error handling to have less unexpected errors

ecooper:
## Problem

We have a few inconsistencies in error handling:
* Account API failures were very verbose and almost always "unexpected errors" even in cases you'd expect them, such as:
  * invalid inputs: like bad region groups
  * 404s: like missing databases
* Common Fauna errors could sneak through and be very confusing. The most notable of this is if you have a bad `--secret`
* Authentication errors in general weren't being consistently caught and displayed.

## Solution

This PR does a few things, but it most notably commits to using the errors defined in `errors.mjs` in our lib code. Through adding new exceptions and common patterns to convert 3rd party errors to things we expect, we can get rid of most unexpected errors.

* New exceptions: Add `AuthenticationError` and `AuthorizationError`. The 401 case has a default message to remove the inconsistent ones we have already.
* `xxxxToCommandError` helpers: Simplified the existing `throwForError` function to just be a translation layer to catch the common error types across our dependencies.

## Result

Authentication errors and 4XX errors display consistently across the account, v4, v10, and schema APIs.

## Testing

Add unit tests for the error mappings. Updated existing error tests.

### DIFF
--- a/src/commands/database/list.mjs
+++ b/src/commands/database/list.mjs
@@ -1,9 +1,7 @@
 //@ts-check
 
-import { FaunaError } from "fauna";
-
 import { container } from "../../cli.mjs";
-import { throwForError } from "../../lib/fauna.mjs";
+import { faunaToCommandError } from "../../lib/fauna.mjs";
 import { FaunaAccountClient } from "../../lib/fauna-account-client.mjs";
 import { colorize, Format } from "../../lib/formatting/colorize.mjs";
 
@@ -61,10 +59,7 @@ async function listDatabasesWithSecret(argv) {
       .resolve("logger")
       .stdout(formatQueryResponse(result, { format: Format.JSON, color }));
   } catch (e) {
-    if (e instanceof FaunaError) {
-      throwForError(e);
-    }
-    throw e;
+    faunaToCommandError(e);
   }
 }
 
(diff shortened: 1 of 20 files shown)


test/schema/schema.mjs
```diff
@@ -35,11 +39,13 @@ describe("schema", function () {
...
-      if (command === "schema push") {
-        return;
-      }
       container.resolve("fetch").rejects(new TypeError("fetch failed"));
+      container.resolve("gatherFSL").resolves([
+        {
+          name: "test.fsl",
+          content: "collection Test { name: String }",
+        },
+      ]);
```
echo-bravo-yahoo, line 48: 10/10, fun with DI testing

echo-bravo-yahoo (approved):

mwilde345 (approved):

src/commands/database/list.mjs
```diff
@@ -50,10 +48,7 @@ async function listDatabasesWithSecret(argv) {
       expression: `Database.all().paginate(${pageSize}).data { ${getOutputFields(argv)} }`,
     });
   } catch (e) {
-    if (e instanceof FaunaError) {
-      throwForError(e);
-    }
-    throw e;
+    faunaToCommandError(e);
```
mwilde345, line 51:
```suggestion
    return faunaToCommandError(e);
```

PR #530, opened 2024-12-16
Add path per line output for db list

ecooper:
## Problem

`fauna db list` only outputs JSON. When running the command without the `--database` flag, it's pretty verbose without providing a ton of value.

## Solution

If we update the output to be a path per line by default, we can simplify the output and shrink it down. `--json` can then output the JSON parsable version.

Because we sometimes use the account API and sometimes use FQL directly, we can only really output a shape that looks like:

```ts
{
  name: string;
  path?: string;
}
```

When `path` is unavailable, we'll output a warning message to stderr to let folks know the full path is not retrievable when using a secret.

## Result

When using `--database`:
Screenshot 2024-12-16 at 10 03 59 AM

When using `--secret`:
Screenshot 2024-12-16 at 9 45 21 AM

## Testing

Updated the tests for list to prove the new output formats.

### DIFF
--- a/src/commands/database/list.mjs
+++ b/src/commands/database/list.mjs
@@ -1,29 +1,10 @@
 //@ts-check
+import chalk from "chalk";
 
 import { container } from "../../cli.mjs";
 import { faunaToCommandError } from "../../lib/fauna.mjs";
 import { FaunaAccountClient } from "../../lib/fauna-account-client.mjs";
 import { colorize, Format } from "../../lib/formatting/colorize.mjs";
-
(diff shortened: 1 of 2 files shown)


cleve-fauna (approved):

mwilde345 (approved):

jrodewig (approved):

jrodewig:
Maybe not in scope here, but we may want to output a message if there are no databases in the current scope. Right now, it's just nothing — which makes sense — but I always wonder if I messed up the command.

PR #536, opened 2024-12-17
Test --include options

ptpaterson:
Ticket(s): FE-###

## Problem
We don't have any tests exercising the --include options.

## Solution

Containerize `formatQueryOptions` and check if it gets called with the right options.

## Result

have tests.

### DIFF
--- a/test/query.mjs
+++ b/test/query.mjs
@@ -6,6 +6,7 @@ import sinon from "sinon";
 
 import { run } from "../src/cli.mjs";
 import { setupTestContainer as setupContainer } from "../src/config/setup-test-container.mjs";
+import { QUERY_INFO_CHOICES } from "../src/lib/command-helpers.mjs";
 import { NETWORK_ERROR_MESSAGE } from "../src/lib/errors.mjs";
 import { colorize } from "../src/lib/formatting/colorize.mjs";
 import {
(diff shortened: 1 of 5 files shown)


test/query.mjs
```diff
@@ -321,9 +322,44 @@ describe("query", function () {
...
-    it.skip("can specify '--include none' to set no include options");
+    it("can set the include option to '[summary]' by default", async function () {
+      const formatQueryInfo = container.resolve("formatQueryInfo");
+
+      await run(`query "foo" --secret=foo`, container);
+
+      expect(formatQueryInfo.getCall(0).args[1].include).to.deep.equal([
+        "summary",
+      ]);
+    });
```
echo-bravo-yahoo, line 333: these tests are a good first start (and fine for now - no reason to re-do this PR), but in general, i think it gives better confidence to test a little deeper. instead of "did my work function receive the right args," you can mock the network call and assert that stderr received the right output - which confirms that _after_ receiving the args, your work function actually did the right thing.

echo-bravo-yahoo (approved):

test/query.mjs
ptpaterson: We do a little bit of that already, just not with different permutations of args.

https://github.com/fauna/fauna-shell/blob/feca5764e2f38ca34722ec3279a5056db36943e4/test/query.mjs#L328-L383

I'll add one more. Easy peasy.

mwilde345 (approved):

PR #539, opened 2024-12-17
Remove --active --no-input flags from fauna local

ecooper:
## Problem

`fauna local` allows you to push a schema while bootstrapping. You need to manually confirm it, and that feels a little weird given the context of the command.

## Solution

Remove `--input` and `--active` from the command, but set `input` to `false` and `active` to `true` when calling `pushSchema`.

## Result

Schema is pushed to the current active schema without a prompt.

## Testing

The existing tests where updated to match the new behavior.


jrodewig (approved):
LGTM.

PR #547, opened 2024-12-18
Clean up lint warnings

ecooper:
## Problem

Over the past few weeks, we've racked up quite a few npm linting warnings.

## Solution

Transfer TODOs into actionable issues, and fix complexity complaints.

## Result

Clean bill of health.

Screenshot 2024-12-18 at 2 16 01 PM

## Testing

No tests have been harmed as part of these changes.

### DIFF
--- a/src/lib/auth/oauth-client.mjs
+++ b/src/lib/auth/oauth-client.mjs
@@ -1,9 +1,18 @@
 import { createHash, randomBytes } from "crypto";
 import http from "http";
 import url from "url";
+import util from "util";
 
 import { container } from "../../cli.mjs";
 import SuccessPage from "./successPage.mjs";
+
+const ALLOWED_ORIGINS = [
+  "http://localhost:3005",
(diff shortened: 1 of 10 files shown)


henryfauna (approved):

src/lib/auth/oauth-client.mjs
```diff
@@ -52,64 +61,73 @@ class OAuthClient {
     return Buffer.from(randomBytes(20)).toString("base64url");
   }
 
+  _handleRedirect({ pathname, res }) {
```
ptpaterson, line 64: Prefix with `#` instead to make it private?

ecooper: We'd need to update the shared eslint configuration and that's outside the scope of this PR.

ptpaterson: Ah, bummer.

PR #548, opened 2024-12-18
Fix colorizing summaries when outputted via --include

ecooper:
## Problem

I noticed summary outputs do not fully colorize the FQL diagnostic help/hints/text when used with `--include`.

Screenshot 2024-12-18 at 3 44 33 PM

Specifically, all "|" and "^" should be red the above example.

## Solution

Update the regular expression used for summary colorization to be the same as query errors. To do this, I needed to strip ansi when checking if it is a fql diagnostic line. This isn't ideal, but it's out of scope to revisit the yaml formatting for these at the moment.

While I was here, I expanded the regex to support lines >9.

## Result

Summaries:
Screenshot 2024-12-18 at 3 44 06 PM

Errors
Screenshot 2024-12-18 at 3 45 02 PM

## Testing

I added a regex test suite for how we detect fql diagnostic lines.

### DIFF
--- a/src/lib/fauna-client.mjs
+++ b/src/lib/fauna-client.mjs
@@ -1,12 +1,18 @@
 //@ts-check
 
+import stripAnsi from "strip-ansi";
+
 import { container } from "../cli.mjs";
 import { isUnknownError } from "./errors.mjs";
 import { faunaToCommandError } from "./fauna.mjs";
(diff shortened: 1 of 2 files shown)


ptpaterson (approved):
Awesome!

henryfauna (approved):

PR #553, opened 2025-01-06
Redact any common secrets from fetch debug logging

ecooper:
## Problem

When outputting to stderr with debugging functions, we do a best effort to redact any secrets printed. We do this to avoid users accidentally copying and pasting debug logging with live secrets from their local terminal to GitHub or other places.

The fetch logging wasn't being redacted.

## Solution

Use the `redactedStringify` helper in the fetch wrapper. While we're here:

* Catch a failure state involving attempting to redact things that aren't strings (this happens mostly in test runs)
* Shorten the '*' fill to be a max of 8 to obfuscate secret length.

## Result

Fetch debug logging will redact anything that we think is a secret:

```zsh
[fetch]: Received 201 of type basic from POST https://account.fauna.com/api/v1/databases/keys with body:
{
  "secret": "fnAF********a-jF",
  "path": "us-std/demo",
  "name": "System generated shell key",
  "role": "admin",
  "ttl": "2025-01-06T18:09:10.034Z",
  "id": "419354064378134593"
}
```

## Testing

Unit tests were updated for redact.


ptpaterson (approved):
Yes please. Ran into this wanting to share debug output with new middleware for scoping secrets.

echo-bravo-yahoo (approved):

PR #556, opened 2025-01-09
Refactor Account API support and container management

ecooper:
## Problem

While working on #554 to support v2 Account API endpoints, a few sharp edges emerged:

* We had no clean way to support v1 vs v2 endpoints
* Account URL was not always read from argv
* It was impossible to modify the DI container outside of `run`, which makes it impossible to test library code directly
* Testing account endpoints required stubbing and testing the fetch layer (vs just stubbing the data you expect)
* Validation and behavior of account endpoints was mixed between `account.mjs` and `fauna-account-client.mjs`.

## Solution

While this could have been at least 2 separate PRs (container changes first, then api client changes), I didn't discover this problem until testing the API changes. If reviewing this is too bothersome, I can break them apart.

The major changes include:
* **Refactor `FaunaAccountClient` class into set of api methods exported as `accountAPI`**: `account.mjs` and `fauna-account-client.mjs` have effectively been merged into `account-api.mjs`. This module contains all the API endpoints we currently support in the CLI as well as helpers to handle Account API responses.
* **Support v1/v2 endpoints**: Requests can now be made to the Account API v1 and v2 endpoints. Errors from v2 endpoints are supported. Endpoint implementations can manage this by passing `version` to the fetch helper in the module.
* **The container singleton is now in `./config/container.mjs`**: In order to test library code directly, we need to be able to set the container outside of `run` contexts. By moving the container singleton into a different module with a setter, `setContainer`, we can easily manage dependency resolution in tests outside of `run` executions.

Places to pay attention:
* `./config/container.mjs`
* `./lib/account-api.mjs`
* `./cli.mjs`
* `./test/mocha-root-hooks.mjs`

Everything else is mostly import changes or test changes.

## Result

You can now call v1 and v2 endpoints. Additionally, the account API interface has been simplified to be:

```ts
import { container } from './config/container.mjs'

// <..snip>
const { listDatabases } = container.resolve('accountAPI')
await listDatabases({ path: 'us-std/demo', pageSize: 10});
```

## Testing

All impacted code paths have either a) new tests or b) updated tests.


src/cli.mjs
```diff
@@ -36,13 +34,14 @@ const __dirname = path.dirname(__filename);
...
+ * @param {import('./config/container.mjs').container} _container - A built and ready for use awilix container with registered injectables.
  */
 export async function run(_argvInput, _container) {
-  container = _container;
+  setContainer(_container);
+
   const argvInput = _argvInput;
-  const logger = container.resolve("logger");
-  const parseYargs = container.resolve("parseYargs");
+  const logger = _container.resolve("logger");
```
echo-bravo-yahoo, line 43: rename to `container`? the declaration these used to conflict with no longer exists

src/lib/utils.mjs
```diff
@@ -18,3 +18,31 @@ export const resolveFormat = (argv) => {
...
   return argv.format;
 };
+
+/**
+ * Standardizes the region of a database path.
+ *
+ * @param {string | undefined} databasePath - The database path to standardize.
+ * @returns {string | undefined} The standardized database path.
+ * @throws {TypeError} If the database path is not a string.
+ */
```
echo-bravo-yahoo, line 28: if this also confirmed the rg was valid, it could save us a lot of distributed correctness checks. thoughts?

echo-bravo-yahoo (approved):

cleve-fauna (approved):

mwilde345 (approved):

PR #558, opened 2025-01-13
FE-6296 Enable scoped database and role with user-provided keys

ptpaterson:
## Problem

You can provide a scoped key by specifying `--secret some_key:child/db:role`, but sometimes it is more convenient to split up the auth config into `--secret some_key -d child/db -r role`. This is especially true if you have a config profile that provides a secret that you sometimes want to scope down, e.g. `-p test-db -d child`.

## Solution

Add middleware that mutates a secret provided with `--secret` (or whatever secret is set if added by prior middleware, such are --local) to scope the key with the provided database and/or role.

In the case of using `--local` or the `local` command, we can remove the special handing of secret, role and database there, since it will be handled for all cases by the new middleware.

## Result

For all commands that accept `--secret`, `--database` and `--role` options can now also be provided.

### Use of credentials files

Using the user-provided secret will now avoid digging through file-based database keys before falling back on the provided secret. To do this, I updated the `getSecret()` function to accept `argv` and short circuit if `argv.secret` already exists. It was brought up that the `getSecret()` function is already a bit of a hack, but I kept this function in place to reduce code duplication: the alternative I saw was to replace `getSecret()` with `argv.secret ?? (await getSecret())` everywhere, and that seemed to add _more_ complexity. I still moved the key-refreshing bit into a class method to keep those details out of the logic that forks at provided key vs credentials file key.

The `Credentials` singletons are still constructed with the container initialization, so we are not avoiding it entirely. Additionally, the check for whether or not to retry 401s with a refreshed key goes through `Credentials` (see `DatabaseKeys.onInvalidCreds()` method). A future optimization might be to not initialize the credentials files with the container, but perform the file I/O when called for and memoize for additional calls. But I was trying to avoid anything that invasive with this PR.

### Example
```bash
❯ ./src/user-entrypoint.mjs query -p test-db -d child --verbosity 5 '42'
[config]: Reading config from /Users/psquared/.fauna/fauna.config.yaml.
[config]: Using profile test-db...
[config]: Applying config: {
    "secret": "fnAF********7Dp5"
}
[config]: Reading config from /Users/psquared/.fauna/fauna.config.yaml.
[config]: Using profile test-db...
[config]: Applying config: {
    "secret": "fnAF********7Dp5"
}
[argv]: {
    "_": [
        "query"
    ],
    "p": "test-db",
    "profile": "test-db",
    "d": "child",
    "database": "child",
    "verbosity": 5,
    "config": "/Users/psquared/.fauna/fauna.config.yaml",
    "secret": "fnAF********7Dp5",
    "color": true,
    "json": false,
    "quiet": false,
    "verbose-component": [],
    "verboseComponent": [],
    "account-url": "https://account.fauna.com",
    "accountUrl": "https://account.fauna.com",
    "user": "default",
    "u": "default",
    "local": false,
    "api-version": "10",
    "v": "10",
    "apiVersion": "10",
    "format": "fql",
    "f": "fql",
    "timeout": 5000,
    "performance-hints": false,
    "performanceHints": false,
    "include": [
        "summary"
    ],
    "$0": "fauna",
    "fql": "42"
}
[argv]: Existing Fauna environment variables: {"FAUNA_CONFIG":"/Users/psquared/.fauna/fauna.config.yaml","FAUNA_PROFILE":"none"}
[argv]: Defaulted url to 'https://db.fauna.com' no --url was provided
[argv]: Applying scope to secret 'fnAF********7Dp5:child:admin', since --database was 'child' with default role 'admin'
[argv]: no --input specified, using [fql]
42
```

## Testing

Added some new tests to cover the `scopeSecret` middleware. This is really just taking over the existing tests that were covered by `applyLocalArg` middleware before.


src/commands/query.mjs
```diff
@@ -87,7 +87,7 @@ async function queryCommand(argv) {
 
   // get the query handler and run the query
   try {
-    const secret = await getSecret();
+    const secret = await getSecret(argv);
```
ecooper, line 90: Is it possible to remove these calls altogether in favor of just mutating argv in a middleware? Then the commands can just use whatever secret is provided and don't need to have a direct dependency on fetching and refreshing dynamic secrets.

ptpaterson: For secrets from credentials, we need to be able to refresh them if they're expired.

Maybe this is okay for all commands except for `shell`?

ptpaterson: I will check again how the secret is updated when a query fails.

ptpaterson: Not all commands need a secret. For example `login`. So we shouldn't add middleware to set a secret from credentials at the top level (`cli.mjs`).

`query`, `shell`, `database`, `schema` are all the commands that accept `--secret`. All of them require the `validateDatabaseOrSecret` check, except for `database list`. `database list` checks for `!argv.secret` and uses the account API if there's no secret.

The way `database list` works at a root level means we need to handle three separate states for the combined group of commands:
- provided secret
- database secret
- no secret

Preserving the "no secret" state seems to be the biggest obstacle to resolving the key thing with a "simple" middleware.

Also, if we are mutating `argv.secret`, we have to make sure to do it _AFTER_ the`validateDatabaseOrSecret` check, which further blocks us from running the middleware at the root command.

Note that `query` and `shell` commands specifically run `validateDatabaseOrSecret` in the handler, rather than a check, with a note that this should give us greater control over errors. I tried moving those to checks, and it didn't break any tests. I don't know what greater control over errors it provides.

# Options

## 1) Leave this as is
Will continue to work okay

## 2) Middleware for commands that need it
We add middleware to set `argv.secret` to the credentials secret. We will end up doing that down in every individual command that needs it at the last moment. It's effectively the same as putting the logic at the start of the commands that need it, so we don't reduce code duplication, but we can move key logic out of the handlers -- the `getSecret(args)` function is converted to a middleware, but it's basically the same.

`query`, `shell`, `database create`, `database delete`, and all `schema` commands all check `validateDatabaseOrSecret`.

We could combine the validation logic and key resolution into a single middleware for these commands.

## 3) Store the secret/key state elsewhere
Separate our concern for what `--secret` is and what we need to use the bearer token for FQL API requests.

Store the key state (provided, credentials, none) in a containerized place. I think Credentials would be a good place. I think this could allow us to optimize DatabaseKeys to not have to care about user-provided keys. The retry on invalid key could still be aware of the key source.

Each command that needs it would then get the secret in the handler. So, replace `const secret = getSecret(argv)` with something like `const secret = credentials.getSecret()`.

We could handle initializing the secret/key state in `buildCredential()` and wouldn't need additional middleware added to each command.

# Proposal

@ecooper I recommend we move forward with this PR to enable the options together, and since anything else is an internal refactor/optimization, I can follow up with something. I wouldn't mind taking a swing at something like option 3 here.

mwilde345 (approved):

PR #559, opened 2025-01-13
FE-6299 Implement FQL formatting for v4 queries

ptpaterson:
## Problem

We are defaulting all queries to JSON, which prints the wire protocol. v10 queries print the FQL format by default.

### Other work while here

Some valid FQL expressions are recognized as executable javascript, including objects and any falsy values like `null` and `false`, so they are either treated strangely or rejected as queries.

## Solution

Output FQL formatting for v4 queries using `util.inspect()`.

Colorize the output using the FQL grammar.

wrap query expressions in an IIFE so that simple values can be provided as queries.

## Result

image

image

## Testing

Added tests to verify formatting and ability to query simple values.


src/lib/formatting/codeToAnsi.mjs
```diff
@@ -66,7 +66,7 @@ const { codeToTokensBase, getTheme } = createHighlighter();
  * Returns a string with ANSI codes applied to the code. This is a JS port of the
  * TypeScript codeToAnsi function from the Shiki library.
  * @param {*} code - The code to format.
- * @param {"fql" | "log" | "json"} language - The language of the code.
+ * @param {"fql" | "log" | "json" | "js"} language - The language of the code.
```
ecooper, line 69: We don't actually have a JS highlighter enabled. Accidentally added?

src/lib/faunadb.mjs
```diff
@@ -46,11 +47,15 @@ export const getClient = async (argv) => {
  */
 export async function stringExpressionToQuery(expression) {
   const faunadb = (await import("faunadb")).default;
-  return runInContext(expression, createContext(faunadb.query));
+
+  const wrappedCode = `(function() { return ${expression} })()`;
```
ecooper, line 51: Could you add a comment around why this is wrapped in anon function call?

ecooper (approved):
Some nits, but otherwise LGTM. Thanks for fixing this.

src/lib/formatting/codeToAnsi.mjs
ptpaterson: I did for a hot minute. I missed this when i reverted it! thanks for the catch.

henryfauna (approved):

PR #573, opened 2025-01-17
Alias --wait on export commands to --watch and -w

ecooper:
## Problem

When using this command, we've seen folks use `--watch` instead of `--wait`.

## Solution

`--wait` is probably the most accurate based on some internal discussions, but we can just support both via aliases. 

## Result

`--wait`, `--watch`, and `-w` will all cause the `export create` and `export get` commands to block until an export hits a terminal state.


cleve-fauna (approved):
i'm ok if you guys just want to use wait too.

jrodewig (approved):

PR #575, opened 2025-01-21
Implement retry options

ptpaterson:
[FE-6286](https://faunadb.atlassian.net/browse/FE-6286)

## Problem

We do not expose the following query options to users:
- v4 queries
  - contention retries
- v10 queries
  - contention retries
  - throttling retry `max_attempts`
  - throttling retry `max_backoff`

The contention retries and throttling retries are slightly separate concerns, but I got confused and ended up with both sets of additional options in this PR.

## Solution

Add the options to the `QUERY_OPTIONS` constant and pipe them into v4 and v10 queries.

The v4 driver does not have an explicit option for configuring the `x-fauna-max-contention-retries` header, but it does accept an object with headers that are included with the query.

## Testing

Added/updated tests to show we're passing the options to the drivers.


ecooper (approved):

src/commands/shell.mjs
```diff
@@ -191,10 +199,13 @@ async function buildCustomEval(argv) {
...
           performanceHints,
           format: outputFormat,
+          maxAttempts,
+          maxBackoff,
+          maxContentionRetries,
         });
 
         // If any query info should be displayed, print to stderr.
-        if (include.length > 0) {
+        if (include && include.length > 0) {
```
echo-bravo-yahoo, line 208: is `include` ever undefined? curious how we'd get into that situation after telling yargs to parse include as an array.

test/commands/shell.mjs
```diff
@@ -338,25 +338,33 @@ describe("shell", function () {
 
     it.skip("does not colorize output if --no-color is used", async function () {});
 
-    it.skip("can eval a query with typechecking enabled", async function () {
-      container.resolve("performV10Query").resolves(v10Object1);
+    it("can open a shell and run queries with options", async function () {
+      runQueryFromString.resolves(v10Object1);
```
echo-bravo-yahoo, line 342: 👍

echo-bravo-yahoo (approved):

PR #578, opened 2025-01-23
[DOCS-4068] Bump Node vers to v22

jrodewig:
## Problem

The `node:sea` module has issues running in Node v20.10.0.

## Solution

Bump the required Node version to v22, the current LTS version.

## Testing

`npm run test:local` is 🟢 .

PR #579, opened 2025-01-23
update package.json engine

echo-bravo-yahoo:
more accurately describe the minimum functional version of node required to run the CLI.

also modifies the tests to run on our minimum supported node version to ensure we don't have regressions in v20.18 that don't show up in v20.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "typescript": "^5.6.3"
   },
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=20.18.0"
   },
   "files": [
     "/dist"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "typescript": "^5.6.3"
   },
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=20.18.0"
```
jrodewig, line 59: This conflicts with https://github.com/fauna/fauna-shell/pull/578. If we decide to go with this one, can we update the README as well?

echo-bravo-yahoo: yes, but i think #578 is a bit too prescriptivist. thoughts on:
```
## Requirements

- [Node.js](https://nodejs.org/en/download/package-manager) v20.18 or later.
  -  [Node.js](https://nodejs.org/en/download/package-manager) v22 or later recommended.
```

ecooper (approved):

jrodewig (approved):

PR #584, opened 2025-01-24
FE-6300 Accept S3 URI as input for export and show S3 URI in output.

cleve-fauna:
## Problem

We got some feedback that exports would be easier to work with if you could input an S3 URI as the desired destination.

As a result our account API was updated to:

1. accept a URI as input
2. always output the destination uri as output.

The output objects now look like:

```
{
   ...rest,
   destination: {
     s3: { bucket: "b", path: "p" },
     uri: "s3://b/p",
   }
}
```

## Solution

This commit updates the CLI :
-  to accept S3 URIs as destination inputs for exports.
- always output the output object that includes a uri field on destionation
- have list use the provided destination.uri rather than constructing one.
## Result

CLI isn't reinventing what the API is already doing.

## Testing

Ran the full suite.
Did some end to end export create, lists and gets with the updated code.

### DIFF
--- a/src/commands/export/create.mjs
+++ b/src/commands/export/create.mjs
@@ -19,19 +19,24 @@ async function createS3Export(argv) {
     wait,
     maxWait,
     quiet,
+    destination,
   } = argv;
   const logger = container.resolve("logger");
   const { createExport } = container.resolve("accountAPI");
-
-  let createdExport = await createExport({
-    database,
-    collections,
-    destination: {
+  let destinationInput = destination;
+  if (!destinationInput) {
+    destinationInput = {
       s3: {
         bucket,
         path,
       },
-    },
+    };
+  }
+
+  let createdExport = await createExport({
+    database,
+    collections,
+    destination: destinationInput,
     format,
   });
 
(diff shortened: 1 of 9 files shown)


echo-bravo-yahoo (approved):

src/commands/export/create.mjs
```diff
@@ -108,7 +124,22 @@ function buildCreateS3ExportCommand(yargs) {
...
         );
       }
-
+      if (argv.destination) {
+        if (argv.bucket || argv.path) {
+          throw new ValidationError(
+            "Cannot specify --destination with --bucket or --path. Use either --destination or both --bucket and --path.",
+          );
+        }
+        if (!S3_URI_REGEX.test(argv.destination)) {
```
ecooper, line 133: Can we either add tests for this regex or tests for the validation function to prove we handle invalid bucket URIs before sending to the API?

ecooper (approved):

src/commands/export/create.mjs
cleve-fauna: sure.

cleve-fauna: Done.

PR #585, opened 2025-01-29
Support idempotency on create export in the CLI.

cleve-fauna:
## Problem

The `POST /v2/exports` API supports idempotency, but this is not exposed in the CLI.

## Solution

Expose it in the CLI for `fauna export create` via an optional `--idempotency` argument

## Result

Customers can create idempotent exports if they desire.

## Testing

Ran the tests also ran with the idempotency command locally against the production API.

You can see that:

- not using `--idempotency` is properly handled (`idempotent_replayed: false` is in the output)
- using it but not replaying is properly handled (`idempotent_replayed: false` is in the output)
- replaying is properly handled  (`idempotent_replayed: true` is in the output)

This proves it works as a steal thread.

Screenshot 2025-01-29 at 9 25 29 AM

### DIFF
--- a/src/commands/export/create.mjs
+++ b/src/commands/export/create.mjs
@@ -154,6 +154,9 @@ function buildCreateS3ExportCommand(yargs) {
           "Either --destination or both --bucket and --path are required to create an export.",
         );
       }
+      if (argv.idempotency?.trim() === "") {
+        throw new ValidationError("--idempotency cannot be an empty string.");
+      }
       return true;
     })
     .example(sharedExamples);
--- a/src/commands/export/create.mjs
+++ b/src/commands/export/create.mjs
@@ -19,6 +19,7 @@ async function createS3Export(argv) {
     maxWait,
     quiet,
     destination,
+    idempotency,
   } = argv;
   const logger = container.resolve("logger");
   const { createExport } = container.resolve("accountAPI");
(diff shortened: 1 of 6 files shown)


src/commands/export/create.mjs
```diff
@@ -116,6 +122,13 @@ function buildCreateS3ExportCommand(yargs) {
         default: "simple",
         group: "API:",
       },
+      idempotency: {
+        type: "string",
+        required: false,
+        description:
+          "Set the idempotency key for the request, prevents replaying the same requests within 24 hours.",
```
jrodewig, line 129:
```suggestion
          "Idempotency key. Avoids reprocessing successful requests with the same key for 24 hours.",
```

src/commands/export/create.mjs
```diff
@@ -140,6 +153,9 @@ function buildCreateS3ExportCommand(yargs) {
           "Either --destination or both --bucket and --path are required to create an export.",
         );
       }
+      if (argv.idempotency?.trim() === "") {
+        throw new ValidationError("--idempotency cannot be an empty string.");
```
jrodewig, line 157:
```suggestion
        throw new ValidationError("--idempotency can't be an empty string.");
```

jrodewig (commented):
Left some nits on the help text. Feel free to disregard as desired.

PR #588, opened 2025-02-05
fix: `--max-backoff` accepts seconds, not ms

jrodewig:
## Problem

The help text and tests for `--max-backoff` flag state/imply it accepts milliseconds.

The JS driver's corresponding `maxBackoff` client config option accepts values in seconds, not milliseconds. The value is converted from seconds to milliseconds here: https://github.com/fauna/fauna-js/blob/main/src/client.ts#L467-L468

Unless I missed it, the CLI does do any conversion of `--max-backoff` values so it should accept seconds.

## Solution

Update the help text and tests to indicate `--max-backoff` accepts seconds.

## Result

Accurate help and tests for the CLI.

## Testing

`npm run test:local` is 🟢

### DIFF
--- a/src/lib/options.mjs
+++ b/src/lib/options.mjs
@@ -143,7 +143,7 @@ export const QUERY_OPTIONS = {
   "max-backoff": {
     type: "number",
     description:
-      "Maximum backoff time (in milliseconds) between retry attempts. Only applies to v10 queries.",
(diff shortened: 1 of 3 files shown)


cleve-fauna (approved):
Yikes, we have some in seconds and some in milliseconds? We need to fix that in the driver(s) and make them all milliseconds.

PR #593, opened 2025-02-11
fix: FQL v4 queries support `--timeout`

jrodewig:
## Problem

The help text for --timeout` state it only applies to FQL v10 queries. However, it is also supported for FQL v4 queries. See this test: https://github.com/fauna/fauna-shell/blob/main/test/commands/query/v4.mjs#L106-L118 

## Solution

Fix the help string. Docs are fixed in https://github.com/fauna/docs.fauna.com/pull/3103

## Result

Accurate help output

## Testing

N/A

### DIFF
--- a/src/lib/options.mjs
+++ b/src/lib/options.mjs
@@ -121,8 +121,7 @@ export const QUERY_OPTIONS = {
   },
   timeout: {
     type: "number",
-    description:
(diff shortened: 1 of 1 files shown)


echo-bravo-yahoo (approved):

ptpaterson (approved):
Great catch!

